PR #19, opened 2019-08-31
Add support for JDKs with JavaFX

### LINKED ISSUE #18: Add support for JavaFX builds
Something like
```
- uses: actions/setup-java@v1
  with:
    java-version: '1.8'
    javafx: true
```

I believe Zulu just has `ca-fx` in the name of jdk versions with JavaFX.

airsquared:
I've tested this on my own repository, and it works well.

I wasn't sure what to set the name of the toolcache to, so [I just changed it to](https://github.com/actions/setup-java/compare/master...airsquared:master#diff-8c84d074e67a54fc5a18692e0b853feaR35) `'fx-jdk'` for JDKs with JavaFX and `'jdk'` for JDKs without JavaFX (you might want to change this to something else).

To use this, just do:

```
- uses: airsquared/setup-java@master
  with:
    java-version: '1.8'
    javafx: true
```

Closes #18.

### DIFF
--- a/__tests__/installer.test.ts
+++ b/__tests__/installer.test.ts
@@ -52,7 +52,7 @@ describe('installer tests', () => {
   }, 100000);
 
   it('Installs version of Java from jdkFile if no matching version is installed', async () => {
-    await installer.getJava('12', 'x64', javaFilePath);
+    await installer.getJava('12', 'x64', javaFilePath, false);
     const JavaDir = path.join(toolDir, 'Java', '12.0.0', 'x64');
 
     expect(fs.existsSync(`${JavaDir}.complete`)).toBe(true);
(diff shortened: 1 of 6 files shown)


rmannibucau:
Hi,

can it be integrated please?

airsquared:
What do you mean by "integrated"?

airsquared:
What do you mean by "integrated"?

rmannibucau:
Merged, it prevents me to use gh actions and i'd like to avoid to fork the download logic again.

giltene:
I've created a separate pull request https://github.com/actions/setup-java/pull/27 that covers the requested feature but with a different syntax than the one proposed here: The change introduces a new java-package option that supports jre, jdk, and jdk+fx as accepted values (defaults to jdk and retains current behavior unless jre or jdk+fx is specified).  I believe that this syntax (a single option to cover current and possible future package types) is preferable to having individual booleans per package type.

The action is available for testing as foo-actions/setup-java@master

konradpabjan:
Support for JavaFx was added with the following PR: https://github.com/actions/setup-java/pull/27 

The README has also been updated with some information about how to use it.

I've also updated the releases tag so as long as you're using `actions/setup-java@v1` then you should just be able to switch one of the parameters to use javafx

PR #20, opened 2019-09-12
JDK 8 specific version help

jameskleeh:
It wasn't clear to me how to specify a version for JDK8 or older. I tried `1.8.152` for example and that doesn't work. `8.0.152` is the correct syntax however that was confusing given `1.8` is supported. This should help those in my situation. If you feel it can be improved please let me know


konradpabjan:
Thanks for the feedback! Good to have some clarity around this.

Resolve the merge conflict and we can get this in 😃

jameskleeh:
@konradpabjan Done

Issue #23, opened 2019-09-29
Not in gzip format


Alexander-Scott:
Hi, I am getting an error when running the setup-java step in the following configuration:
```yaml
    - uses: actions/setup-java@v1
      with:
        java-version: '12.x'
```

The error I get is the following:

```console
Run actions/setup-java@v1
/bin/tar xzC /home/runner/work/_temp/temp_1966052730 -f /home/runner/work/_temp/7c187b07-8201-4bf2-b87d-247013d1c81d

gzip: stdin: not in gzip format
/bin/tar: Child returned status 1
/bin/tar: Error is not recoverable: exiting now
##[error]The process '/bin/tar' failed with exit code 2
##[error]Node run failed with exit code 1
```

After doing some googling, [this](https://stackoverflow.com/questions/16381917/issues-uncompressing-a-tar-gz-file) looks to be the most relatable problem, but I notice it's not being downloaded the Oracle site, rather from this [site](https://cdn.azul.com/zulu/bin/) so I don't think cookies are the issue. Has anyone else also encountered this error?

bijancn:
I also see this issue. Here for example https://github.com/starship/starship/pull/383/checks?check_run_id=240449201

pzygielo:
Experiencing the same for `java-version: 1.8` as well.

ray26:
I have got this error too in Java-version:1.8

zhzy0077:
Having the same problem. It was fine several hours ago.

ConorTheDev:
Also having this issue on my [repo](https://github.com/MediaModMC/MediaMod)

joschi:
Same problem here with Java 1.8, 11, and 13.

dropwizard/dropwizard#2945

amardeshbd:
Getting same error in 

* https://github.com/amardeshbd/android-barebone/commit/53ccf2935b70c4de9ef03a83faebd033d677bbad/checks
* https://github.com/amardeshbd/android-barebone/commit/bfa508fd067c00716b9eb2b400b5fd3165a34baa/checks

```
2019-09-29T13:48:22.8414796Z ##[group]Run actions/setup-java@v1
2019-09-29T13:48:22.8414968Z with:
2019-09-29T13:48:22.8415075Z   java-version: 1.8
2019-09-29T13:48:22.8415178Z   architecture: x64
2019-09-29T13:48:22.8415261Z ##[endgroup]
[...]

PR #24, opened 2019-09-30
Match filename extension only at  the end of the filename

giltene:
Changes the matching for filename extensions to only match extensions at the end of the filename, and ignore false matches with the same pattern in the middle of the name. This avoids wrongly matching .tar.gz.sig and zip.sig files in the CDN.

The change will prevent issues like https://github.com/actions/setup-java/issues/23 happening in the future if/as .sig files are added on the CDN. For the time being, the Zulu CDN has been cleansed of .sig files to avoid this issue, but we would like to be able to safely .sig files back up once this change makes it into a 1.x version of setup-java...

I have verified this change the entire range of Java versions (6, 7, 8, 9, 10, 11, 12, 13), update levels, and OS platforms available on the CDN at this point in time.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Set Node.js 10.x
       uses: actions/setup-node@master
       with:
-        version: 10.x
+        node-version: 10.x
 
     - name: npm install
       run: npm install
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -7,6 +7,8 @@ jobs:
     strategy:
       matrix:
         operating-system: [ubuntu-latest, windows-latest]
+      fail-fast: false
+      max-parallel: 5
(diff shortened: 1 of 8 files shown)


giltene:
Ping...

jawher:
A workaround until this PR is merged is to use the @giltene's fork:

```yml
- name: Set up JDK 11
  uses: giltene/setup-java@master
  with:
    java-version: 11
```

giltene:
@damccorm, @stephenmichaelf can one of you review this change and hopefully merge it? We would really like to put .sig files back where they belong on the Zulu cdn, and this is what is blocking that right now...

damccorm:
@bryanmacfarlane could you triage appropriately?

bryanmacfarlane:
@madhurig

giltene:
PR https://github.com/actions/setup-java/pull/27 includes this change with a cleaner history, so I’m closing this one.

PR #27, opened 2019-11-03
Add java-package parameter to action, support jre, jdk, and jdk+fx

giltene:
This change adds support for multiple Java package types: jre, jdk, and jdk+fx. In addition to the [updated] built-in tests, this change has been tested against a wide range of Java versions and package type combinations that exist on the Zulu community cdn, verifying that existing jdk (default) behavior remains the same, and new package types (jre, jdk+fx) work as well on versions for which they are available in the cdn.

* Add java-package option to action, support jre, jdk, and jdk+fx (#1)

* Update tests to use 'jdk', 'jre', and 'jdk+fx' java package option, as well as test with invalid package 

* Match extension only at end of line (e.g. ignore .tar.gz.sig files in cdn).

* Update README.md

* Update workflow to use 'node-version' instead of deprecated 'version'

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 This action sets up a java environment for use in actions by:
 
-- optionally downloading and caching a version of java by version and adding to PATH. Downloads from [Azul's Zulu distribution](http://static.azul.com/zulu/bin/).
+- optionally downloading and caching a requested version of java by version and adding to PATH. Default downloads are populated from the [Zulu Community distribution of OpenJDK](http://static.azul.com/zulu/bin/)
 - registering problem matchers for error output
 
 # Usage
@@ -19,7 +19,8 @@ steps:
 - uses: actions/checkout@latest
 - uses: actions/setup-java@v1
   with:
-    java-version: '9.0.4' // The JDK version to make available on the path. Takes a whole or semver Jdk version, or 1.x syntax (e.g. 1.8 => Jdk 8.x)
+    java-version: '11.0.4' // The Java version to make available on the path. Takes a whole or semver Java version, or 1.x syntax (e.g. 1.8 => Jdk 8.x)
+    java-package: jdk // (jre, jdk, or jdk+fx) - defaults to jdk
     architecture: x64 // (x64 or x86) - defaults to x64
 - run: java -cp java HelloWorldApp
 ```
@@ -32,7 +33,7 @@ steps:
   with:
     java-version: '4.0.0'
     architecture: x64
-    jdkFile: <path to jdkFile> # Optional - jdkFile to install java from. Useful for versions not supported by Azul
+    jdkFile: <path to jdkFile> # Optional - jdkFile to install java from. Useful for versions not found on Zulu Community CDN
 - run: java -cp java HelloWorldApp
 ```
 
@@ -43,7 +44,8 @@ jobs:
     runs-on: ubuntu-16.04
     strategy:
       matrix:
-        java: [ '1.6', '9.0.x', '12.0.2' ]
+        # test against latest update of each major Java version, as well as specific updates of LTS versions:
+        java: [ 1.6, 6.0.83, 7, 7.0.181, 8, 8.0.192, 9.0,x, 10, 11.0.x, 11.0.3, 12, 13 ]
     name: Java ${{ matrix.java }} sample
     steps:
       - uses: actions/checkout@master
@@ -51,7 +53,6 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: ${{ matrix.java }}
-          architecture: x64
       - run: java -cp java HelloWorldApp
 ```
 
--- a/action.yml
+++ b/action.yml
@@ -3,10 +3,14 @@ description: 'Setup your runner with Java'
 author: 'GitHub'
 inputs: 
   java-version:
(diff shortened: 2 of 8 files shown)


lib/installer.js
```diff
@@ -176,22 +176,39 @@ function getDownloadInfo(refs, version) {
...
+        pkgTypeLength = 'fx-jdk'.length;
+    }
+    else {
+        throw new Error(`package argument ${javaPackage} is not in [jdk | jre | jdk+fx]`);
+    }
     // Maps version to url
     let versionMap = new Map();
     // Filter by platform
     refs.forEach(ref => {
-        if (ref.indexOf(extension) < 0) {
```
giltene: Fixes https://github.com/actions/setup-java/issues/23

giltene:
Note that in addition to adding the option to specify  jre and jdk-fx packages addressing https://github.com/actions/setup-java/issues/18, the change to match extension names only at the end of line fur file name (makes sure to ignore e.g. .tar.gz.sig) addresses the things that have caused https://github.com/actions/setup-java/issues/23 , so that .sig files can be safely added to the cdn without breaking actions/setup-java

action.yml
```diff
@@ -3,10 +3,14 @@ description: 'Setup your runner with Java'
 author: 'GitHub'
 inputs: 
   java-version:
-    description: 'The JDK version to make available on the path. Takes a whole or semver Jdk version, or 1.x syntax (e.g. 1.8 => Jdk 8.x)'
+    description: 'The Java version to make available on the path. Takes a whole or semver Java version, or 1.x syntax (e.g. 1.8 => Jdk 8.x)'
     required: true
+  java-package:
+    description: 'The package type (jre, jdk, jdk+fx)'
+    required: true
```
konradpabjan: Could you change this `false`, same thing for the architecture. They have default versions so they don't need to be  explicitly included in the YAML.

I tested it out with my own fork and it looks to be working as expected if you don't provide them. This is good because it won't break any existing workflows that use this action.
https://github.com/konradpabjan/Testing3/blob/master/.github/workflows/java.yml
https://github.com/konradpabjan/setup-java

I overall really like the direction of this. If there are any other packages in the future that users would like then it makes it easy to add it them here and no more extra input parameters will have to be added

README.md
```diff
@@ -19,7 +19,8 @@ steps:
 - uses: actions/checkout@latest
```
konradpabjan: Not really related to this PR, but could you change it from `actions/checkout@latest` to `actions/checkout@v1`. The build fails if you use `latest`. Want to have the YAML working correctly if people copy and paste it.

konradpabjan: Disregard. There was a another PR that fixed this. Some merge conflicts though that will have to be resolved.

action.yml
giltene: I was wondering about that. I chose to mimick the way the existing architecture input is defined, and for that one required is set to true, but the input provides a default. If we change one we should probably change both, and for consistency, we would also need to change the way they are consumed in e.g. https://github.com/foo-actions/setup-java/blob/master/src/setup-java.ts#L12 and https://github.com/foo-actions/setup-java/blob/master/lib/setup-java.js#L29

konradpabjan: If an input type has a default value then I'm leaning towards setting its required type to `false`. I think of what is in `action.yml` as the interface for whoever is using the action and separate from whatever is in `src/` consuming it. If an input has a default value and is false, then it makes it abundantly clear that the end user doesn't have to actually provide anything when using the action for that input type and they can use something like:

```
- uses: actions/setup-java@v1
  with:
    java-version: '9.0.4'
```

When an input is attempted to be read/consumed, its required type is set to false by default. In our case we have it set to `true` which will result in an exception being thrown if nothing is provided which is what we want to happen. https://github.com/actions/toolkit/blob/master/docs/specs/package-specs.md

Behind the scenes, a default value that we specify in `action.yml` creates an environment variable on the runner: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/metadata-syntax-for-github-actions#inputs

If there is some error on the GitHub back-end that causes environment variables to not be created (highly unlikely) then an exception being thrown when trying to read the default input env variable will help identify that it's not an end-user error.

giltene: I'm fine with either way for this, but should we change the architecture option the same way at the same time? (It seems wrong to have one do it one way and the other do it the other way)

konradpabjan: Yes, architecture as well (knock them both out), so it should be:
```
java-package:
    description: 'The package type (jre, jdk, jdk+fx)'
    required: false
    default: 'jdk'
architecture:
    description: 'The architecture (x86, x64) of the package.'
    required: false
    default: 'x64'
```

giltene: Ok. And to be clear, leave https://github.com/foo-actions/setup-java/blob/master/src/setup-java.ts#L12 as it is?
Or do we need to change:
    `const javaPackage = core.getInput('java-package', {required: true});`
to something like:
    `const javaPackage = core.getInput('java-package', {required: false}) || 'jdk';`
?

konradpabjan: Yes, leave everything in `setup-java.ts` as is. The only changes are going to be in `action.yml`

lib/installer.js
giltene: The change in already incorporated into this PR

konradpabjan:
Thanks for getting these changes in! 😃

PR #29, opened 2019-11-16
Adding maven auth support

clarkbw:
Hello!

I wasn't sure about best practices as I'm working up this PR to enable publishing for Maven.

### Problem

Publishing via Apache Maven requires a `settings.xml` file. ([see docs](https://maven.apache.org/settings.html)) which can have 2 locations:

- The Maven install: `${maven.home}/conf/settings.xml`
- A user’s install: `${user.home}/.m2/settings.xml`

Where the user install would override any config from the default Maven install.

#### Option 1 (my plan)

So my plan is to create the file `$HOME/.m2/settings.xml` with the following contents.

```xml
<settings>
    <servers>
    <server>
        <username>clarkbw</username>
        <password>$GITHUB_TOKEN</password>
    </server>
    </servers>
</settings>
```

Where the username and password come from new config options passed into this action.

```yaml
steps:
- uses: actions/checkout@master
- uses: actions/setup-java@v1
  with:
    java-version: '4.0.0'
    architecture: x64
    username: clarkbw
    password: $GITHUB_TOKEN
- run: mvn publish
```

#### Option 2

There is an alternate method people use where you create [reference variables](https://stackoverflow.com/questions/28071697/is-it-possible-to-pass-a-password-in-maven-deploy-in-the-command-line) in the settings file and then pass your options in on the command line.

`settings.xml` file looks like this:

```xml
<settings>
    <servers>
    <server>
        <username>${repo.login}</username>
        <password>${repo.pwd}</password>
    </server>
    </servers>
</settings>
<settings>
```

In this setup there are no new config options, instead you pass in a set of command line options such that the CLI looks like this: `mvn -Drepo.login=clarkbw -Drepo.pwd=$GITHUB_TOKEN publish`

I don't like this because we end up requiring these new magic `repo.login` and `repo.pwd` command line options instead of some new YAML config options.

### Additional notes

**Alternate Apache Maven Home**

It is possible to create an alternate Apache Maven home.  If the `M2_HOME` env variable is set I could (possibly, assuming it is writable) write the file `$M2_HOME/conf/settings.xml` but it usually isn't user writable and can have other useful defaults we'd lose.

**Maven Options for option 2**
It looks like it is possible to write / append to the `MAVEN_OPTS` env variable with the `-Drepo.login=clarkbw -Drepo.pwd=$GITHUB_TOKEN` CL options such that the user isn't required to specify them on the command line.  However this actually is passed over to the JVM and could have some unintended consequences.  

### Question

What are the best practices for writing the `$HOME/.m2/settings.xml` file?

I essentially did something like this:

```javascript
const dir = path.join(os.homedir(), '.m2');
io.mkdirP(dir);
const settings = path.join(dir, 'settings.xml');
fs.writeFileSync(settings, XML);
```

Is writing to the $HOME directory ok?  Are there better env variables to use?


lib/setup-java.js
```diff
@@ -28,6 +29,11 @@ function run() {
             const arch = core.getInput('architecture', { required: true });
             const jdkFile = core.getInput('jdkFile', { required: false }) || '';
             yield installer.getJava(version, arch, jdkFile);
+            const username = core.getInput('username', { required: false });
+            const password = core.getInput('password', { required: false });
+            if (username && password) {
+                yield auth.configAuthentication(username, password);
+            }
```
clarkbw: This could be `maven-username` and `maven-password` to make it clearer.

src/auth.ts
```diff
@@ -0,0 +1,31 @@
...
+export async function configAuthentication(username: string, password: string) {
+  const directory: string = path.join(os.homedir(), '.m2');
+  await io.mkdirP(directory);
+  await write(directory, generate(username, password));
+}
+
+// only exported for testing purposes
+export function generate(
+  username = '${actions.username}',
+  password = '${actions.password}'
```
clarkbw: I did this to allow for the option 2 as a possible fallback but would need to make changes to the `setup-java` such that undefined parameters are passed.  If this route looks reasonable I can make the changes and users could pass in CL options like I described in option 2.  If its extra noise I'm happy to remove it.

clarkbw: Removed this so there isn't confusion.

giltene:
Something about putting maven-specific things into a generic setup-java action seems wrong. Can you elaborate on what is wrong with the "option 2" solution to the problem? That one seems to work without needing anything to change in setup-java, and without introducing username/password options to the action that have no actual use within it (just to propagate them to later users). If we really need that, perhaps it should be in a separate setup-username-and-password action?

konradpabjan:
Initially I was thinking this should go into a separate action but we want packaging to work on work out of the box, and Maven is the de-facto package ecosystem for Java. This can also be configured to work with `Gradle` since they use the same registry. 

In option 1, you would also need to update `action.yml` to specify the input parameters and provide them with descriptions and set them as not-required. Having `username` and `password` as extra input is fine as long as it's optional and doesn't break existing workflows. I'm leaning towards this approach since it's easier to use and doesn't require an extra CLI step to be run before `setup-java`. It has to be optional though.

Could you update `README.md` with some example of YAML of how this could be used

src/setup-java.ts
```diff
@@ -13,6 +14,11 @@ async function run() {
 
     await installer.getJava(version, arch, jdkFile);
 
+    const username = core.getInput('username', {required: false});
+    const password = core.getInput('password', {required: false});
+
+    await auth.configAuthentication(username, password);
```
konradpabjan: I would add a `if` statement here early on to avoid even jumping into the method if one of the two inputs is not provided

src/auth.ts
```diff
@@ -0,0 +1,43 @@
...
+import * as io from '@actions/io';
+
+export const M2_DIR = '.m2';
+export const SETTINGS_FILE = 'settings.xml';
+
+export async function configAuthentication(username: string, password: string) {
+  if (username && password) {
+    core.debug(`configAuthentication with ${username} and a password`);
+    const directory: string = path.join(os.homedir(), M2_DIR);
+    await io.mkdirP(directory);
```
konradpabjan: `io.mkdirP` works fine to create a new directory 👍

src/auth.ts
```diff
@@ -0,0 +1,43 @@
...
+    core.debug(`created directory ${directory}`);
```
konradpabjan: `core.debug` is fine. By we don't necessarily want users to have to enable step debugging to see anything. Certain things can be outputted normally.

src/setup-java.ts
clarkbw: Ok, I moved it down lower.  

I wrapped the call in the control statement as requested.  I don't like this because its now an untested path. I have tests within the auth module for input validation because its simpler to test the smaller module than this file.

src/auth.ts
clarkbw: Ok, I'll adjust some of the output to include `console.log` output where it would be more generally useful.

clarkbw:
Ok, I've finished my updates.  

You can see a successful run of this here: https://github.com/pied-piper-inc/fib-apache-maven/commit/0048dcb399de48db4bf184e7c1b1aa24f7eabeeb/checks?check_suite_id=333934912

I might add 1 additional note to the README about publishing to other registries.

giltene:
Tons of people out there use gradle, and we'd probably want to make sure this seems "symmetric" from a build tool perspective. The actual change here will [presumably] work with both, and is not maven-tool-specific. To avoid making this appear maven-tool-specific, I would:
1. Add a working gradle example to match the maven one, with both using the server-id, username, and password fields. [e.g. under "Publishing to an Apache Maven Repository:" in the README you can have separate "with maven:" and "with gradle:" examples].
2. Make sure any reference to maven (e.g. in field descriptions in action.yml) unambiguously refers to Apache Maven repositories, rather than using looser terms like "maven" or "maven server".
3. Probably even change the title from "Adding maven auth support" to something like "Adding auth support fields for convenience in publishing to Apache maven repositories via e.g. maven & gradle"

README.md
```diff
@@ -56,6 +56,38 @@ jobs:
...
+```yaml
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@master
```
konradpabjan:
```suggestion
      uses: actions/setup-java@v1
```

README.md
```diff
@@ -56,6 +56,38 @@ jobs:
...
+        java-version: 1.8
+        server-id: github # Value of the distributionManagement/repository/id field of the pom.xml
+        username: ${{ github.actor }} # username for server authentication
+        password: ${{ github.token }} # password or token for authentication
+    - name: Build with Maven
+      run: mvn -B package --file pom.xml
+    - name: Publish to GitHub Packages Apache Maven
+      run: mvn deploy
+    - name: Set up Apache Maven Central
+      uses: actions/setup-java@master
```
konradpabjan:
```suggestion
      uses: actions/setup-java@v1
```

src/auth.ts
```diff
@@ -0,0 +1,50 @@
...
+export async function configAuthentication(
+  id: string,
+  username: string,
+  password: string
+) {
+  if (id && username && password) {
+    console.log(
+      `creating ${SETTINGS_FILE} with server-id: ${id}, username: ${username}, and a password`
+    );
+    const directory: string = path.join(os.homedir(), M2_DIR);
```
konradpabjan: Didn't mention this earlier, but the `$GITHUB_WORKSPACE` directory is where we should be creating any new directories or files. Whenever we use `actions/checkout`, any checked out code gets saved to to the directory so we should mimic the same behavior: https://github.com/actions/checkout

This does also brings up the potential scenario of what-if someone already has a .`m2` with the same settings file. I would add an extra try-catch block around `fs.writeFileSync`. There should be a clear error message and a failure if we're unable to create the necessary files. You can use `@actions/core` to do that: https://github.com/actions/toolkit/tree/master/packages/core

README.md
```diff
@@ -56,6 +56,38 @@ jobs:
...
+      with: # running setup-java again overwrites the settings.xml
+        java-version: 1.8
+        server-id: maven
+        username: maven_username
+        password: ${{ secrets.MAVEN_CENTRAL_TOKEN }} # password from secrets store
+    - name: Publish to Apache Maven Central
```
konradpabjan: Could you also add an example using Gradle

src/auth.ts
clarkbw: > Didn't mention this earlier, but the `$GITHUB_WORKSPACE` directory is where we should be creating any new directories or files. Whenever we use `actions/checkout`, any checked out code gets saved to to the directory so we should mimic the same behavior: https://github.com/actions/checkout

I'll give this a try.  The `settings.xml` file needs to live in `~/.m2/` so if `$GITHUB_WORKSPACE` isn't the same as `$HOME` this might not work. 



> This does bring up the potential scenario of what-if someone already has a .`m2` directory with the same settings file if they've used checkout. You do mention in the example in the `README` that running the action again causes an overwrite to happen. I think it would be good to detect this behavior and output some message to inform the user about this overwrite.

Good idea, I'll add something to check for an existing file and log if we're overwriting it.

README.md
clarkbw: I was going to do Gradle in a follow up PR. Lemme try to wrap this up and if I can I'll add an example as well.

src/auth.ts
clarkbw: > I'll give this a try. The `settings.xml` file needs to live in `~/.m2/` so if `$GITHUB_WORKSPACE` isn't the same as `$HOME` this might not work.

This writes to the wrong directory I'm seeing in the logs:
```
writing /home/runner/work/fib-apache-maven/fib-apache-maven/.m2/settings.xml
```

And then the authentication fails.  See https://github.com/pied-piper-inc/fib-apache-maven/commit/19b690b1b6c87e9a7aa468dba112e7f546b13a3f/checks?check_suite_id=343056828

clarkbw: Did some more investigation and I can override the setting with the command line flag `-s $GITHUB_WORKSPACE/.m2/settings.xml` which makes the full deploy `mvn deploy -s $GITHUB_WORKSPACE/.m2/settings.xml`

Its not as nice as the straight `mvn deploy` but if the `$GITHUB_WORKSPACE` is a better place than `$HOME` for these kind of auth files we can do that instead.  Lemme know what you think.

konradpabjan: The end goal is to have a simple experience, and an extra `-s` flag complicates the whole scenario which we don't want to do.

Lets keep it in `$HOME`. I would only just add a small note in the `README` and clarify that this is exactly where the `settings.xml` file will be created.

README.md
konradpabjan: Another PR with a Gradle example is fine. As long as we eventually have one.

konradpabjan:
Small nit: A few trailing spaces in `action.yml`. See the failing `Lint YAML` action.

I should add this linting so it can be spotted easily during development (something like `npm run-script lint`), but that's more work for me in the future

src/auth.ts
clarkbw: Ok, done!

clarkbw:
Successful publish https://github.com/pied-piper-inc/fib-apache-maven/commit/ce5f751c905488f5125b840499423c73dc1e0ee0/checks?check_suite_id=346312038#step:5:283

action.yml
```diff
@@ -14,7 +16,19 @@ inputs:
...
+    required: false
+  server-id:
+    description: 'ID of the distributionManagement repository in the pom.xml
+       file.'
+    required: false
+  username:
+    description: 'Username for authentication to the Apache Maven server.'
+    required: false
+  password:
+    description: 'Passowrd or token for authentication to the Apache Maven
```
konradpabjan:
```suggestion
    description: 'Password or token for authentication to the Apache Maven
```

action.yml
```diff
@@ -14,7 +16,19 @@ inputs:
...
+       server.'
```
konradpabjan:
```suggestion
       repository.'
```
Per the suggestion of @giltene

action.yml
```diff
@@ -14,7 +16,19 @@ inputs:
...
-    description: 'Path to where the compressed JDK is located. The path could be in your source repository or a local path on the agent.'
+    description: 'Path to where the compressed JDK is located. The path could
+       be in your source repository or a local path on the agent.'
+    required: false
+  server-id:
+    description: 'ID of the distributionManagement repository in the pom.xml
+       file.'
+    required: false
+  username:
+    description: 'Username for authentication to the Apache Maven server.'
```
konradpabjan:
```suggestion
    description: 'Username for authentication to the Apache Maven repository.'
```

clarkbw:
Woo!

README.md
clarkbw: Ok, added a Gradle example in 203550b
I have a demo build of gradle here https://github.com/pied-piper-inc/fib-gradle/commit/2b7855275aaa4153ba7f9fe64b3e9ef3b8272817/checks?check_suite_id=346412071

README.md
```diff
@@ -56,6 +56,68 @@ jobs:
...
+      run: mvn deploy 
+```
+
+***NOTE: The `settings.xml` is created in the Actons $HOME directory by default and will overwrite existing files.***
```
konradpabjan:
```suggestion
***NOTE: The `settings.xml` file is created in the Actions $HOME directory. If you have an existing `settings.xml` file at that location, it will be overwritten***	
```
Small spelling fix. Also some rewording. Sounds a little scary right now as it seems like all existing files will be overwritten.

README.md
```diff
@@ -56,6 +56,68 @@ jobs:
...
+        password: ${{ secrets.MAVEN_CENTRAL_TOKEN }} # password from secrets store
+    - name: Publish to Apache Maven Central
+      run: mvn deploy 
+```
+
+***NOTE: The `settings.xml` is created in the Actons $HOME directory by default and will overwrite existing files.***	
+
+See the help docs on [Publishing a Package](https://help.github.com/en/github/managing-packages-with-github-packages/configuring-apache-maven-for-use-with-github-packages#publishing-a-package) for more information on the `pom.xml` file.
+
+Publishing using Gradle:
```
konradpabjan: Good example 👍 Thanks for adding this!

infin8x:
One note here... I _think_ (but am not certain) that on Actions self-hosted runners, `$HOME` is shared by all the actions running on a machine. So, putting anything in the home directory means that you might accidentally leak credentials from one run to another. @chrispat is my memory up to date here?

clarkbw:
> Actions self-hosted runners, `$HOME` is shared by all the actions running on a machine.

Yes, this is correct.  There aren't many options available to us.  I went through what I found here: https://github.com/actions/setup-java/pull/29#discussion_r353819014

I have been wondering if I can detect being run within a shared runner.  I'd be happy to push out a warning in that case.  Otherwise the console log will let people know we are writing or overwriting into `$HOME`

konradpabjan:
Good catch @infin8x There is an env variable called `GITHUB_ACTIONS` that you can use to detect this scenario: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables

chrispat:
While I hate extra options I think we should have an option to put the file in some other location and then output an environment variable that points to the settings file so the user can reference it.  I am fine with the default just working for the hosted runners by putting the file in home but for self-hosted we are going to need an option.

src/auth.ts
```diff
@@ -0,0 +1,61 @@
...
+    await write(directory, generate(id, username, password));
+  } else {
+    core.debug(
+      `no ${SETTINGS_FILE} without server-id: ${id}, username: ${username}, and a password`
+    );
+  }
+}
+
+// only exported for testing purposes
+export function generate(id: string, username: string, password: string) {
```
JLLeitschuh: You're not sanitizing XML control characters here.
You need to replace XML control characters with their 'safe' counterparts.

https://stackoverflow.com/a/7918944/12344388

clarkbw: Good catch! fixed this in 551e2a2

clarkbw:
> Good catch @infin8x There is an env variable called `GITHUB_ACTIONS` that you can use to detect this scenario

That looks like I can detect local vs. Actions, is that the same as a shared runner?

__tests__/auth.test.ts
```diff
@@ -82,4 +82,22 @@ describe('auth tests', () => {
...
+      <servers>
+        <server>
+          <id>${id}</id>
+          <username>${username}</username>
+          <password>&amp;&lt;&gt;&quot;&apos;&apos;&quot;&gt;&lt;&amp;</password>
+        </server>
+      </servers>
+  </settings>
+  `);
+  });
```
JLLeitschuh: Woot! 🎆

clarkbw:
Ok, as noted in the new docs https://github.com/clarkbw/setup-java/tree/maven-auth#apache-maven-within-a-shared-runner You can now specify an alternate location when using a shared runner but in your `mvn deploy` you need to specify the location of the settings file. 

In the first pass at the docs I kept the `.m2` directory but realized later it is confusing to tack that on when a user has specified a directory.

README.md
```diff
@@ -56,6 +56,94 @@ jobs:
...
+      env:
+        USERNAME: ${{ github.actor }}
+        PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+```
+
+***NOTE: The `USERNAME` and `PASSWORD` need to correspond to the credentials environment variables used in the publishing section of your `build.gradle`..***	
+
+See the help docs on [Publishing a Package with Gradle](https://help.github.com/en/github/managing-packages-with-github-packages/configuring-gradle-for-use-with-github-packages#example-using-gradle-groovy-for-a-single-package-in-a-repository) for more information on the `build.gradle` configuration file.
+
+## Apache Maven within a Shared Runner
```
chrispat: Shared runner is not really the right term here.  It would be a self-hosted runner with multiple runners configured under the same user on the same machine.

README.md
```diff
@@ -56,6 +56,94 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8 for Shared Runner
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+        server-id: github # Value of the distributionManagement/repository/id field of the pom.xml
+        username: ${{ github.actor }} # username for server authentication
+        password: ${{ github.token }} # password or token for authentication
+        m2-home: ${{ $GITHUB_WORKSPACE }} # location for the settings.xml file
```
chrispat: I don't see `m2-home` in the `action.yml` below.

clarkbw: Yeah, I switch that to the settings name and add it to the yaml file. 

AFAIK It is not possible to use the `$GITHUB_WORKSPACE` env to set this path. The only thing that works is using relative or absolute paths like `./config` or `/config/` respectively.

chrispat: In your example above it would need to be `${{ github.workspace }}` . You can't reference a bash environment variable in the macro.

clarkbw:
Updated the docs with suggestions from above, new examples seen here: https://github.com/clarkbw/setup-java/tree/maven-auth#apache-maven-with-a-settings-path

konradpabjan:
Looks good to me!. @chrispat could you take one more look before I merge the changes in

action.yml
```diff
@@ -14,7 +16,22 @@ inputs:
...
+    required: false
+  username:
+    description: 'Username for authentication to the Apache Maven repository.'
+    required: false
+  password:
+    description: 'Password or token for authentication to the Apache Maven
+       repository.'
+    required: false
+  settings-path:
+    description: 'Path to where the settings.xml file will be written.'
```
chrispat:
```suggestion
    description: 'Path to where the settings.xml file will be written. Default is ~/.m2.'
```

src/setup-java.ts
```diff
@@ -16,6 +17,14 @@ async function run() {
 
     const matchersPath = path.join(__dirname, '..', '.github');
     console.log(`##[add-matcher]${path.join(matchersPath, 'java.json')}`);
+
+    const id = core.getInput('server-id', {required: false});
+    const username = core.getInput('username', {required: false});
+    const password = core.getInput('password', {required: false});
```
joshmgross: You might want to set `password` as a secret to protect users from accidental password leakage (in case they're populating this input without a SECRET already). This will enable masking for logs.

```ts
core.setSecret(password);
```
See https://github.com/actions/toolkit/tree/master/packages/core#setting-a-secret

chrispat: If they are storing the password in the file unencrypted it makes no difference if it is masked in the logs.

src/auth.ts
```diff
@@ -0,0 +1,75 @@
...
+async function write(directory: string, settings: string) {
+  const options = {encoding: 'utf-8', flag: 'wx'}; // 'wx': Like 'w' but fails if path exists
+  const location = path.join(directory, SETTINGS_FILE);
+  console.log(`writing ${location}`);
+  try {
+    return fs.writeFileSync(location, settings, options);
+  } catch (e) {
+    if (e.code == 'EEXIST') {
+      console.warn(`overwriting existing file ${location}`);
+      return fs.writeFileSync(location, settings, {
```
brcrista: As a rule of thumb, exceptions shouldn't be used for control flow. I think a better way to do this would be to check if the file exists first:

```ts
async function write(directory: string, settings: string) {
  const location = path.join(directory, SETTINGS_FILE);

  if (fs.accessSync(location, fs.constants.F_OK)) {
      console.warn(`overwriting existing file ${location}`);
  } else {
      console.log(`writing ${location}`);
  }

  return fs.writeFileSync(location, settings, {
      encoding: 'utf-8',
      flag: 'w'
  });
}
```

You can also modify this to check if the file exists but is read-only, but I think throwing an exception is fine in that case.

action.yml
```diff
@@ -14,7 +16,22 @@ inputs:
...
   jdkFile:
-    description: 'Path to where the compressed JDK is located. The path could be in your source repository or a local path on the agent.'
+    description: 'Path to where the compressed JDK is located. The path could
+       be in your source repository or a local path on the agent.'
+    required: false
+  server-id:
+    description: 'ID of the distributionManagement repository in the pom.xml
+       file.'
+    required: false
+  username:
```
brcrista: Would it be appropriate to call these `maven-username` or `server-username`? I think it makes the YAML read more naturally:

```yml
- name: Set up JDK 1.8
  uses: actions/setup-java@v1
  with:
    java-version: 1.8
    server-id: github
    server-username: ${{ github.actor }}
    server-password: ${{ github.token }}
```

src/setup-java.ts
brcrista: Should we print a warning if a user sets one of these, but doesn't set them all?

package.json
```diff
@@ -8,7 +8,7 @@
     "build": "tsc",
     "format": "prettier --write **/*.ts",
     "format-check": "prettier --check **/*.ts",
-    "release": "ncc build && git add -f dist/",
+    "release": "ncc build lib/setup-java.js && git add -f dist/index.js",
```
joshmgross: Why have `ncc` build the javascript files? It can handle typescript

konradpabjan: Nice catch! The `.ts` source file can be used. See the following example: https://github.com/zeit/ncc/blob/bfefbbce9f5bdf5ba9514fd1eb0c133aa3ba3d32/examples/typescript/package.json#L8

action.yml
konradpabjan: I like `server-username` for it's consistency, makes it obvious that all 3 variables are related to each other

README.md
```diff
@@ -56,6 +56,94 @@ jobs:
...
+
+## Publishing using Gradle
+```yaml
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
```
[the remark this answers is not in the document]
clarkbw: I'll do some general spacing for clarity  and uniformity but I don't think a space between every step is required.

action.yml
clarkbw: Going with `server-*`

src/setup-java.ts
clarkbw: Will do

src/auth.ts
clarkbw: Thanks for the example! I used `existsSync` which is effectively the same `accessSync` but can be used in control statements.

https://nodejs.org/api/fs.html#fs_fs_accesssync_path_mode
https://nodejs.org/api/fs.html#fs_fs_existssync_path

src/setup-java.ts
clarkbw: And... someone only now shared with me that we can use env variables within the `settings.xml` file. https://stackoverflow.com/a/31251518

So a generic settings file like so might work and we won't need to save the token into it.

```xml
<servers>
    <server>
      <id>github</id>
    </server>
</servers>
```

clarkbw: I did this but then with the new env variable system I removed this as all inputs are totally optional now.

package.json
clarkbw: Updated this.  Please double check as I'm new to `ncc`

clarkbw:
Ok, everything is updated again now.

The new code is using env variables in the `settings.xml` file.  You can now have a very simple setup-java where you pass the `$GITHUB_TOKEN` to deploy and everything else works (assuming `github` is the ID you use in your `pom.xml` file).

```yaml
    steps:
    - uses: actions/checkout@v1
    - name: Set up JDK 1.8
      uses: actions/setup-java@v1
      with:
        java-version: 1.8

    - name: Build with Maven
      run: mvn -B package --file pom.xml

    - name: Publish to GitHub Packages Apache Maven
      run: mvn deploy
      env:
        GITHUB_TOKEN: ${{ github.token }} # GITHUB_TOKEN is the default env for the password
```

The new `settings.xml` file generated no longer contains passwords.  We still need to allow people to configure the ID of the server as that corresponds to their `pom.xml`.  The id defaults to `github`, username defaults to `$GITHUB_ACTOR`, and password defaults to `$GITHUB_TOKEN`.  

```xml
<servers>
    <server>
      <id>github</id>
    </server>
</servers>
```

Each of these environment variable names are configurable.   I did this so the default above can be very simple and when using an alternative like Maven people don't try to set the GITHUB_ACTOR env.  Another option would have been to use something like `$MAVEN_USER` and `$MAVEN_PASSWORD` env variables that aren't configurable but that requires more env configuration in the simplest case.  I'm happy to drop this if consensus is against the configuration.

ccleve:
I'm the person who posted about Github documentation problems on Hacker News a few hours ago. Thank you for your response.

I'll write more tomorrow, but what we really need is a complete, working example of a canonical Hello World Java/Maven app that can be built on Actions and deployed on Packages. I suggest creating a new Github project that does this. I should be able to fork it and run it in my own account without modification. You may want to create more than one such project to handle different situations.

Put copious notes right in the POM, settings, and .yml explaining exactly what is going on.

Mention this and similar projects (for npm, Gradle, etc.) prominently right in the written documentation. The docs can be a walkthrough of the sample app, with an explanation of what is going on and why.

Then provide reference docs. Have a doc that shows, in detail, what all of the options are in the action .yml. You also need a reference on all the environment variables that are available. I hadn't heard of GITHUB_ACTOR until I saw it above. Where is that documented?

Having app(s) like this will give us a place to post issues and bugs.

Be sure to read these:
https://stackoverflow.com/questions/59147676/error-code-400-when-trying-to-maven-deploy-to-github-packages-using-github-act
https://stackoverflow.com/questions/59313581/how-to-install-a-package-hosted-by-github-packages-with-github-actions?noredirect=1#comment105020411_59313581

Also, I suggest not putting settings.xml in the normal ~/.m2 dir. Put it in the root of the project, and then call "mvn --settings settings.xml deploy". That keeps everything in one place without the need to copy settings around.

If you find that getting the maven-deploy-plugin is too much of a pain, I suggest abandoning it and writing your own Maven plugin. A maven-github-packages plugin would avoid all of this configuration hell. All configuration could go right in the plugin config without the need for anything external. No settings.xml, no repositories, no distributionManagement.

package.json
konradpabjan: What you have is works and I tested it out. There are quite a few ways to configure ncc in `package.json`. Relatively new to `ncc` as well 😁

 Might change it up a little bit to align it with some of our other actions that also use `ncc` like `actions/cache` but I'll handle that myself in another PR. Thanks!

konradpabjan:
Thanks @clarkbw for the hard work on this PR! Looks good to me so I'll be merging this into `master`. 🎉

I'll hold off till Monday before merging this into `releases/v1` and at that point anyone using `actions/setup-java@v1` will be able to use this. Holding off since it's a Friday and I want to do some extra testing before getting the changes hit everyone.

BigBlueHundids (commented):
.>

Issue #34, opened 2019-12-22
Add/improve EA version support


giltene:
The current version specification logic in setup-java allows us to "spell" e.g. 14.0.0-ea, but semver version matching and comparison rules are "less than ideal". There are three key problems with the current logic:
1. It does not correctly match to the latest EA build: Semver will use build numbers (in the version forms of e.g. 14.0.0-ea+b28) in precedence comparisons. As a result, the current spelling will match to the first (not latest) ea build in the CDN
- Note: we've started hiding older EA build filenames from the index in the CDN to keep this working for people for now, but     this is not a good practice or one we want to keep doing.
2. It does not allow the specification of a specific EA version (e.g. EA b27 vs EA b28). Semver will ignore build numbers (spelled with e.g. +b28), and match an explicitly requested e.g. 14.0.0-ea+b28 to a cdn 14.0.0-ea+b29
3. It does not support pre-release version formats that would allow #1 and #2 above to be resolved by changing the CDN file name to fit the existing logic. While the CDN can be populated with e.g. 14.0.0-ea.28, 14.0.0-ea will not match that. And will translated the current logic 14.0.0-ea.28 to 14.0.0-ea.28.x, which will not match (semver does not support .x syntax for pre-release version matching).

To prepare for an improvement, we have started populating the CDN with EA version number forms that semver can validly match and do comparisons on (e.g. 14.0.0-ea.28, 15.0.0-ea.2), but the current logic will not match to those. I will be issuing a pull request (shortly) with suggested logic updates that will correctly work with those new version number forms.= and solve the problems noted above.

giltene:
Resolved with the iteration of https://github.com/actions/setup-java/pull/35 into https://github.com/actions/setup-java/releases/tag/v1.3.0

Thanks!

PR #35, opened 2019-12-22
Add more robust support for EA version specification

giltene:
These  changes will address https://github.com/actions/setup-java/issues/34 and add support specifying EA versions in the following format examples. E.g.:
14-ea
14.0.0-ea
14.0.0-ea.28

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -4643,12 +4643,21 @@ function normalizeVersion(version) {
             throw new Error('1. is not a valid version');
         }
     }
-    // Add trailing .x if it is missing
-    if (version.split('.').length != 3) {
+    if (version.endsWith('-ea')) {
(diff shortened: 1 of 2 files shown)


giltene:
Agreed on the README, but I'd like to add the README examples separately, after we get out a version that would allow the CDN can hold more than one EA build at a time. The reason is that in the README, I would want to show how to specify a specific EA build version (e.g. 14.0.0-ea.27 vs. 14.0.0-ea.28), but for now, to get around the issue this pull request fixes, we've been "hiding" the older versions on the CDN such that only the one version appears at a time.
Once we get the new logic in, we'll stop hiding older EA versions in the CDN, and then a README example requesting a specific (non-latest) EA build will work right. I'd want to do a separate pull request for the README at that point.

PR #41, opened 2020-01-09
Update REAME and action.yml to describe java-version syntax options

giltene:
Show how to spell EA versions, as well as more readily describe options for version specification in README


sormuras:
Tested `15-ea` here: https://github.com/sormuras/sandbox/runs/383180220

PR #42, opened 2020-01-17
Add support for maven/gradle versions

### LINKED ISSUE #40: Allow to customize Maven version
It should be possible to set a specific Maven version to use with `mvn`. For that it should be optionally possible to download also Maven from some Apache mirror (https://maven.apache.org/download.cgi) after downloading Java and then use that mvn version.

jvalkeal:
- Essentially use same concepts as jdk install to switch
  maven and gradle versions if those are defined with config.
- Bump to tool-cache 1.3.0 as 1.0.0 tries to use
  node_modules/@actions/tool-cache/scripts/externals/unzip which doesn't have
  execute flag so you'd get `Permission denied`
- `extractZipNix` in toolkit no longer support relative path so needs to
  resolve absolute path before passing file to these methods.
- Commit has my local build for `index.js` which probably should get
  recreated but makes it easier to test this from a PR branch.
- Fixes #40

For example this:
```
    - name: Set up JDK 1.8
      uses: jvalkeal/setup-java@gh40
      with:
        java-version: 1.8
        maven-version: 3.6.1
        gradle-version: 5.6.2

    - name: maven home
      run: echo $M2_HOME

    - name: maven version
      run: mvn -v

    - name: gradle home
      run: echo $GRADLE_HOME

    - name: gradle version
      run: gradle -v
```

Will output something like:
```
2020-01-17T13:00:16.5204913Z /opt/hostedtoolcache/maven/3.6.1/x64
...
2020-01-17T13:00:17.2252202Z Apache Maven 3.6.1 (d66c9c0b3152b2e69ee9bac180bb8fcc8e6af555; 2019-04-04T19:00:29Z)
...
2020-01-17T13:00:17.2965824Z /opt/hostedtoolcache/gradle/5.6.2/x64
...
2020-01-17T13:00:17.8507839Z Welcome to Gradle 5.6.2!
2020-01-17T13:00:17.8507925Z 
2020-01-17T13:00:17.8508036Z Here are the highlights of this release:
2020-01-17T13:00:17.8511449Z  - Incremental Groovy compilation
2020-01-17T13:00:17.8520303Z  - Groovy compile avoidance
2020-01-17T13:00:17.8520773Z  - Test fixtures for Java projects
2020-01-17T13:00:17.8521043Z  - Manage plugin versions via settings script
2020-01-17T13:00:17.8521091Z 
2020-01-17T13:00:17.8521370Z For more details see https://docs.gradle.org/5.6.2/release-notes.html
2020-01-17T13:00:17.8521457Z 
2020-01-17T13:00:17.8521499Z 
2020-01-17T13:00:17.8521724Z ------------------------------------------------------------
2020-01-17T13:00:17.8521821Z Gradle 5.6.2
2020-01-17T13:00:17.8522042Z ------------------------------------------------------------
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,22 @@ steps:
 - uses: actions/setup-java@v1
   with:
     java-version: '4.0.0'
-    architecture: x64
     jdkFile: <path to jdkFile> # Optional - jdkFile to install java from. Useful for versions not found on Zulu Community CDN
+- run: java -cp java HelloWorldApp
+```
+
+## Maven/Gradle Versions
+```yaml
+steps:
+- uses: actions/checkout@v1
+- uses: actions/setup-java@v1
+  with:
+    java-version: '9.0.4'
+    maven-version: '3.6.2'
+    maven-mirror: <uri to maven mirror directory> # Optional - defaults to https://archive.apache.org/dist/maven/maven-3/
+    maven-file: <path to maven-file> # Optional - to install maven from.
+    gradle-version: '5.6.2'
+    gradle-file: <path to gradle-file> # Optional - to install gradle from.
 - run: java -cp java HelloWorldApp
 ```
 
@@ -87,7 +101,7 @@ jobs:
         server-password: MAVEN_CENTRAL_TOKEN # env variable for token in deploy
 
     - name: Publish to Apache Maven Central
-      run: mvn deploy 
+      run: mvn deploy
       env:
         MAVEN_USERNAME: maven_username123
         MAVEN_CENTRAL_TOKEN: ${{ secrets.MAVEN_CENTRAL_TOKEN }}
@@ -117,7 +131,7 @@ The two `settings.xml` files created from the above example look like the follow
 </servers>
 ```
 
-***NOTE: The `settings.xml` file is created in the Actions $HOME directory. If you have an existing `settings.xml` file at that location, it will be overwritten. See below for using the `settings-path` to change your `settings.xml` file location.***	
+***NOTE: The `settings.xml` file is created in the Actions $HOME directory. If you have an existing `settings.xml` file at that location, it will be overwritten. See below for using the `settings-path` to change your `settings.xml` file location.***
 
 See the help docs on [Publishing a Package](https://help.github.com/en/github/managing-packages-with-github-packages/configuring-apache-maven-for-use-with-github-packages#publishing-a-package) for more information on the `pom.xml` file.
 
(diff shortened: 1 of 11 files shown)


martinpaljak:
Would be nice to have this! Maybe the PR has better chance if it is rebased to current master?

joschi:
@jvalkeal Why not use the [Gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) or [Maven wrapper](https://github.com/takari/maven-wrapper) instead?

That way you can build the project locally and aren't depending on the GitHub action to set up everything for you.

jvalkeal:
@joschi I do use wrappers but this was just a way to pick maven/gradle versions if you don't use wrappers.

joschi:
@jvalkeal I'm afraid that this would lead to feature creep in this GitHub action.

Why stop at Gradle and Maven, why not add SBT, Leiningen, Buildr, Ivy, and Bazel, too? And these are just the build systems shown at https://search.maven.org/. 😉

There are dedicated GitHub actions for setting up [Gradle](https://github.com/marketplace/actions/gradle-command) and [Maven](https://github.com/marketplace/actions/setup-maven), so I wonder if it makes sense to include it in `setup-java`.

brunoborges:
I agree this proposal goes way beyond the objective of this action and should be dropped, given all that has been said already.

kwin:
Then you should get rid of all Maven artifacts in setup-java. Currently it already sets up the server.xml in https://github.com/actions/setup-java/blob/4003c04fbcd7c6d5d67fa11e57d706021663f6b2/src/auth.ts#L44. But I agree that probably separating set up of Java and Build System makes sense.

joschi:
@kwin Not disagreeing with you here. In my humble opinion this kind of functionality should live in the respective GitHub actions for Maven and Gradle.

Neither @brunoborges nor I are contributors to or maintainers of this GitHub action, so we can only offer our input. 😉

stCarolas:
@joshmgross @clarkbw @konradpabjan What do you think about discussion above and issue https://github.com/actions/setup-java/issues/60 ?

clarkbw:
I've seen this question come up several times for different actions. And we should likely clarify the contribution guidelines for this action so it is clear.  We are not interested in supporting all clients.  These actions exist as the best way to get ~80% of customers up and running quickly.  This means good defaults and we support the majority clients / configurations.  Supporting all configurations and clients would make this too complicated to maintain and use.

For setup-java I think that means we support maven and gradle out of the box as these are very common client configurations.  Of course we won't support all of the maven or gradle features and so we should direct people to those more complete actions for their areas.  Similar for Ivy and anything else we should likely link out to known good actions.  Its going to be hard to know if this feature of maven or gradle is required to be supported so it might be a judgement call sometimes.

dmitry-shibanov:
Hello everyone. I'm closing this pull request, because we don't have plans to add support for maven or gradle to setup-java, as it was discussed in https://github.com/actions/setup-java/issues/40.

PR #49, opened 2020-02-11
standardize workflow test pattern

ericsciple:


.github/workflows/workflow.yml
```diff
@@ -1,57 +1,85 @@
...
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
         operating-system: [ubuntu-latest, windows-latest]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
-
-    - name: Setup Node.js 12.x
+    - name: Setup Node.js 12
```
ericsciple: i removed the empty lines between steps since the file is much larger now (tighten up individual jobs, to make file more skimmable)

.github/workflows/workflow.yml
```diff
@@ -1,57 +1,85 @@
...
-    services:
-      squid:
-        image: dakale/squid
-        ports: ['3128:3128']
-        options: '--health-cmd "exit 0" --health-interval 3s'
-    env:
-      http_proxy: http://localhost:3128
-      https_proxy: http://localhost:3128
+
+  test:
```
ericsciple: added E2E test job that leverages `./`

.github/workflows/workflow.yml
```diff
@@ -1,57 +1,85 @@
...
+    - name: Verify Java 13 (Windows)
+      if: runner.os == 'windows'
+      run: __tests__/verify-java.ps1 13.0.2
 
-    - name: npm install
-      run: npm install
-
-    - name: Lint
-      run: npm run format-check
+  test-proxy:
```
ericsciple: Switched the proxy test to:
1. Run in a container, with dns disabled (`options: --dns 127.0.0.`). This guarantees that all requests are actually going through the proxy. Previous approach hacked hosts file which does not guarantee all traffic.
2. Switched to widely used squid container image, rather than maintain our own
3. Switched to testing `./` during proxy test rather than unit tests

.github/workflows/workflow.yml
```diff
@@ -1,57 +1,85 @@
...
+    - name: Setup Java 13
+      uses: ./
+      with:
+        java-version: 13.0.2
+    - name: Verify Java 13
+      run: __tests__/verify-java.sh 13.0.2
 
-    - name: npm test
-      run: npm test
+  test-bypass-proxy:
```
ericsciple: Added an E2E proxy bypass test also

__tests__/verify-java.ps1
```diff
@@ -0,0 +1,11 @@
+if (!$args.Count -or !$args[0])
```
ericsciple: helper script to validate a specific version of java is in the PATH

giltene:
> LGTM 👍
> 
> Really good additions that will help with testing & validation.

+1

 > Testing only one version of Java (the latest, version 13) seems sufficient. No point in kicking off a huge Matrix that is effectively doing the same thing considering there are a lot of versions of Java that we support


FYI, I've been keeping an eye on things externally with a pretty full matrix here: https://github.com/giltene/jdkTestingMatrix to make sure the various versions and spellings (e.g. ea, non-ea). Not sure if we need this to be part of the project or if external monitoring is sufficient, but it does serve as an interesting status if/when people observe problems and want to check to see if it is a systemic issue.

PR #53, opened 2020-04-12
Add 'latest' as alias for version 'x'

plamentotev:
With the new release cycle a new JDK version is released once every six months. It may not seems like that often, but still adds overhead to update the build matrix with the latest JDK version. It is more convenient to set the build to run on the LTS versions that the project supports plus the latest JDK version. I've created a [proof of concept](https://github.com/plamentotev/plexus-archiver/pull/2/checks?check_run_id=580910207) workflow [1]. 

While it is possible to specify that you want the latest JDK version by using semver X-Ranges (like 'x' for example), 'latest' is more readable as explicitly states the intention. Also if you use the value from the build matrix in your job name, "JDK:latest" is better than "JDK:x".

 [1] The workflow uses my change as asked by the contributors guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ Examples of version specifications that the java-version parameter will accept:
   
   e.g. ```8.0.x, >11.0.3, >=13.0.1, <8.0.212```
   
+  e.g. ```latest``` (same as ```x```)
+
 - An early access (EA) Java version
 
   e.g. ```14-ea, 15-ea```
--- a/dist/index.js
+++ b/dist/index.js
@@ -4873,6 +4873,11 @@ function getDownloadInfo(refs, version, javaPackage) {
     return { version: curVersion, url: curUrl };
 }
 function normalizeVersion(version) {
+    // 'latest' means the most recent of any JDK version
+    // semver won't match pre-release versions
+    if (version === 'latest') {
+        version = 'x';
+    }
     if (version.slice(0, 2) === '1.') {
         // Trim leading 1. for versions like 1.8
         version = version.slice(2);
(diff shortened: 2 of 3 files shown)


plamentotev:
To be honest I'm not quite sure if the tools cache plays well with 'x' as version (or any other semver range for that matter. e.g  `>=13.0.1`)

giltene:
> To be honest I'm not quite sure if the tools cache plays well with `x` as version (or any other semver range for that matter. e.g `>=13.0.1`)

Since the semver version parsing is done directly on the CDN contents, the caching (outside of the CDN itself) is not a significant issue. My experience has been that things we put in the CDN show up in .x version results fairly quickly (the longest I've seen had been about 30 minutes).

giltene:
+1 for this. It took me a while to understand what it is for (since x technically works), but since the string we use to specify the version in a matrix is then typically used in the job name, I agree that seeing "Java latest" as a job name is much more readable than "Java x".

plamentotev:
Actually I did some tests. If 'x' is specified the version in the tools cache is used. But if 'latest' is specified the version in tools cache is not found as `tc.find` is not aware of the alias. I wonder if we should first normalize the version and then search in the tool cache. I'll check how it works with `1.x` and `-ea` versions.

plamentotev:
Yes, it does not work for 1.8 as well. I'll open an issue and provide PR.

maxim-lobanov:
Hello @plamentotev , if it is still actual for you - we don't have objections to accept this PR.
But it will require pulling latest main, we have recently released V2 so your changes should be adopted for it: https://github.com/actions/setup-java/blob/main/src/distributions/base-installer.ts#L120

plamentotev:
I think I found a better way to achieve the same so I'll close it.

PR #56, opened 2020-04-14
Normalize the version before it is looked up in the tool cache

### LINKED ISSUE #54: 1.x versions does not work with the tools cache
When Java package is added to the cache it is added using the package version. For example 8.0.181. When a Java package is looked up in the tool cache the version supplied by the user is used. That could be either 8 or 1.8. Both should match 8.0.181, but as you can expect `tc.find` won't match 1.8.

plamentotev:
Currently the resolved version is cached. For example something like 8.0.141. On other hand when a version is looked up in the cache the string provided by the user is used. This may lead to unexpected behavior. 1.8 will not match 8.0.141. Normalizing the version string before looking up in the tool cache will solve the problem. The normalized version (8.x) will match the entry in the cache as expected.

Normalizing the string in the beginning of the flow won't hurt as it is a valid version specification. What is more it the user can supply the normalized version directly so the code should be able to work with it.

Fixes #54

### DIFF
--- a/__tests__/installer.test.ts
+++ b/__tests__/installer.test.ts
@@ -143,6 +143,20 @@ describe('installer tests', () => {
     return;
   });
 
+  it('Uses 1.x version of Java installed in cache', async () => {
+    const JavaDir: string = path.join(toolDir, 'jdk', '8.0.181', 'x64');
+    await io.mkdirP(JavaDir);
+    fs.writeFileSync(`${JavaDir}.complete`, 'hello');
+    // This will throw if it doesn't find it in the cache
+    await installer.getJava(
+      '1.8',
+      'x64',
+      'path shouldnt matter, found in cache',
+      'jdk'
+    );
+    return;
+  });
+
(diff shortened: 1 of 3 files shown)


plamentotev:
Judging by the failed test it is not the resolved version that is being stored in the cache but the version spec (the version string provided by the user.)

plamentotev:
Hmm, I can update my change so that the version is normalized only if jdkFile is not provided. Then it will work. 

Or at least kind of. It will work if the user entered version like 12 and not 12.x On one hand it makes sense - why would you enter version range if you supplied the file. But on other hand does it makes sense to use version and tool cache if you have supplied the jdk file? What would be the use case? To populate the cache for other jobs that does not provide jdkFile? Because if I provide the file then does it matter what version I provide and why would I like to take it from the cache?

ZacSweers:
Is this PR being looked by anyone at github?

maxim-lobanov:
Closing according https://github.com/actions/setup-java/issues/54#issuecomment-814773590

PR #57, opened 2020-04-15
Support 32-bit architecture

### LINKED ISSUE #55: `architecture` parameter is ignored
I'm trying to install an x86 JDK in a GitHub action, and I found out that the specified architecture is ignored in this script. By looking at the code, I saw that `arch` is not used when the script is determining which archive to download. See https://github.com/actions/setup-java/blob/5c87b70ffefc9ca2e77153b3528c6b7bc85a84a9/src/installer.ts#L194-L203 On Windows, it will always download the archive ending with `-win_x64`, while it should actually download the archive ending with `-win_i686.zip` when the architecture is `x86`. The issue also seems to be present for other OSes (although I never tried them)

Be aware that Zulu doesn't seem to provide x86 Windows binaries for all Java versions.

emgre:
Fix #55 

I'm using this code in this GitHub Actions pipeline: https://github.com/dnp3/opendnp3/actions

I only tried it on Windows. Perhaps some integration tests could be added to verify that the proper version is fetched. I'm not sure to understand how your tests are run, so I didn't do it.

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -60,7 +60,7 @@ export async function getJava(
 
       const contents = await response.readBody();
       const refs = contents.match(/<a href.*\">/gi) || [];
-      const downloadInfo = getDownloadInfo(refs, version, javaPackage);
+      const downloadInfo = getDownloadInfo(refs, version, arch, javaPackage);
       jdkFile = await tc.downloadTool(downloadInfo.url);
       version = downloadInfo.version;
       compressedFileExtension = IS_WINDOWS ? '.zip' : '.tar.gz';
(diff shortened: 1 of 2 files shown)


src/installer.ts
```diff
@@ -188,20 +188,33 @@ async function unzipJavaDownload(
...
   refs: string[],
   version: string,
+  arch: string,
   javaPackage: string
 ): {version: string; url: string} {
   version = normalizeVersion(version);
+
+  let archExtension = '';
+  if (arch === 'x86') {
+    archExtension = 'i686';
```
joshmgross, line 198: Where did this extension come from? Is it not `x86`?

src/installer.ts
```diff
@@ -188,20 +188,33 @@ async function unzipJavaDownload(
...
+  } else if (arch === 'x64') {
+    archExtension = 'x64';
+  } else {
+    throw new Error(`architecture "${arch}" is not int [x86 | x64]`);
```
joshmgross, line 202: We should validate that the architecture input is valid when we retrieve the input in https://github.com/actions/setup-java/blob/master/src/setup-java.ts#L12

Then this whole block can just be:
```
const archExtension = arch === 'x86' ? 'i686' : 'x64';
```

src/installer.ts
emgre: `i686` is the latest iteration of the `IA-32` (commonly known as `x86`). From today's perspective, it's a synonym of `x86` (although it is technically not). I didn't choose this extension, I just looked at https://cdn.azul.com/zulu/bin/ and saw that's what they use.

src/installer.ts
emgre: Indeed. I'm not very familiar with the GitHub Actions core library (and JS in general) so bear with me. I'll make the change soon.

joshmgross: No worries! Let me know if you need any help

AustinShalit:
Hi folks. Any progress updates on this PR?

src/installer.ts
PeterJohnson: Should be "not in" instead of "not int".

AustinShalit:
@konradpabjan This PR can now be closed.

joshmgross:
Replaced with https://github.com/actions/setup-java/pull/95

Issue #43, opened 2020-01-18
Support GPG


jaredpetersen:
Releasing artifacts to Maven Central requires publishers to sign all of the assets with GPG. Publishers most commonly deal with this by utilizing [maven-gpg-plugin](https://maven.apache.org/plugins/maven-gpg-plugin/).

This is a bit of a headache right now in GitHub actions as you have to create a step in your workflow that:
1. takes a GitHub secret containing your private key and writes it to a file:
  `echo $GPG_PRIVATE_KEY > private.asc`
2. imports the gpg key:
  `gpg --import --batch private.asc`

And to utilize the GPG key, you then have to pass it in as a maven argument:
`-Dgpg.passphrase=$GPG_PASSPHRASE`

This is definitely workable (see [link](https://github.com/jaredpetersen/actions-testbed/blob/f23733ce957e28ed0b329a3a6d5f5b96b7720253/.github/workflows/release.yml#L15-L28) for an example workflow) but it's more tedious than needed. It's also not ideal to pass passwords around using the command line even though GitHub Actions sanitizes them. Since setup-java already creates a settings.xml file that is preconfigured to publish to Maven Central, it would be nice if gpg could also be added to the settings.xml file.

The usage would look something like the following:
```
- name: Set up Apache Maven Central
  uses: actions/setup-java@v1
  with: # running setup-java again overwrites the settings.xml
     java-version: 1.8
     server-id: maven # Value of the distributionManagement/repository/id field of the pom.xml
     server-username: MAVEN_USERNAME # env variable for username in deploy
     server-password: MAVEN_CENTRAL_TOKEN # env variable for token in deploy
     gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }} # gpg private key to import
     gpg-passphrase: GPG_PASSPHRASE # env variable for gpg signing in deploy
- name: Publish to Apache Maven Central
  run: mvn deploy 
  env:
    MAVEN_USERNAME: maven_username123
    MAVEN_CENTRAL_TOKEN: ${{ secrets.MAVEN_CENTRAL_TOKEN }}
    GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
```

And the generated settings.xml would look something like:
```xml
<servers>
    <server>
      <id>maven</id>
      <username>${env.MAVEN_USERNAME}</username>
      <password>${env.MAVEN_CENTRAL_TOKEN}</password>
    </server>
    <profiles>
      <profile>
        <activation>
[...]

PR #59, opened 2020-05-02
Add support for GPG

jaredpetersen:
Adds support for GPG per #43.

I still need to demonstrate a successful test run per the contributor docs, but I figured it would be useful to get this up now.


jaredpetersen:
Two successful matrix test runs:
- [Actions Run - Default values](https://github.com/jaredpetersen/setup-java-gpg/actions/runs/94249962)
- [Actions Run - Override values](https://github.com/jaredpetersen/setup-java-gpg/actions/runs/94249961)

You can view the `List contents` step to show the signed files.

__tests__/auth.test.ts
```diff
@@ -138,7 +158,50 @@ describe('auth tests', () => {
           <password>\${env.&amp;&lt;&gt;&quot;&apos;&apos;&quot;&gt;&lt;&amp;}</password>
         </server>
       </servers>
+      <profiles>
+        <profile>
+          <activation>
+            <activeByDefault>true</activeByDefault>
+          </activation>
+          <properties>
+            <gpg.passphrase>\${env.${gpgPassphrase}}</gpg.passphrase>
```
kwin, line 167: putting this in a server section is more in line with the recommended setup at http://maven.apache.org/plugins/maven-gpg-plugin/usage.html#Configure_passphrase_in_settings.xml and also later on would allow for storing it in an encrypted fashion. Having it in a default profile as property is just too much overhead IMHO.

jaredpetersen:
**Updated**

Two successful matrix test runs:
- [Actions Run - Default values](https://github.com/jaredpetersen/setup-java-gpg/actions/runs/94550674)
- [Actions Run - Override values](https://github.com/jaredpetersen/setup-java-gpg/actions/runs/94550675)

View the `Output settings` step to show the generated `settings.xml` file.

View the `List contents` step to show the signed files.

jaredpetersen:
@joshmgross Can you take a look at this PR or let me know who the appropriate reviewer should be? Thanks!

joshmgross:
Thanks for the ping @jaredpetersen, I'll take a look at this today.

src/auth.ts
```diff
@@ -3,32 +3,59 @@ import * as os from 'os';
...
 export const M2_DIR = '.m2';
 export const SETTINGS_FILE = 'settings.xml';
+export const PRIVATE_KEY_DIR = '.keys';
+export const PRIVATE_KEY_FILE = 'private-key.asc';
 
 export const DEFAULT_ID = 'github';
 export const DEFAULT_USERNAME = 'GITHUB_ACTOR';
 export const DEFAULT_PASSWORD = 'GITHUB_TOKEN';
+export const DEFAULT_GPG_PRIVATE_KEY = '';
+export const DEFAULT_GPG_PASSPHRASE = 'GPG_PASSPHRASE';
```
joshmgross, line 17: Should we default to an empty string here?

src/auth.ts
```diff
@@ -54,21 +82,34 @@ export function generate(
           <username>\${env.${escapeXML(username)}}</username>
           <password>\${env.${escapeXML(password)}}</password>
         </server>
+        <server>
```
joshmgross, line 85: If a user doesn't specify a GPG key, won't this be empty?

Will that cause issues for users who don't want GPG authentication

joshmgross:
@clarkbw would you mind taking a look at this?

src/auth.ts
jaredpetersen: This does default to an empty string. Just like GITHUB_TOKEN and GITHUB_ACTOR are environment variables, GPG_PASSPHRASE is an environment variable as well.

src/auth.ts
jaredpetersen: Users will see the following by default whether they want GPG authentication or not:
```
<settings>
    <servers>
      <server>
        <id>github</id>
      </server>
      <server>
        <id>gpg.passphrase</id>
        <passphrase>${env.GPG_PASSPHRASE}</passphrase>
      </server>
    </servers>
</settings>
```

And GPG_PASSPHRASE won't be provided so it will be empty string. This isn't a problem for users that don't want GPG -- maven will behave normally and will only grab information from this other server config if it needs it for the GPG plugin.

__tests__/auth.test.ts
```diff
@@ -128,17 +147,55 @@ describe('auth tests', () => {
...
+  }, 100000);
+
+  it('does not import gpg private key when private key is not set', async () => {
+    const id = 'packages';
+    const username = 'USERNAME';
+    const password = 'PASSWORD';
+
+    await auth.configAuthentication(id, username, password);
+
+    expect(exec.exec).not.toHaveBeenCalledWith(
```
clarkbw, line 193: I’d make this less precise and go with toHaveBeenCalled instead. If your exec call drifts and for example the cwd is passed incorrectly this would still pass.

src/auth.ts
```diff
@@ -3,32 +3,59 @@ import * as os from 'os';
...
+  await write(
+    settingsDirectory,
+    SETTINGS_FILE,
+    generate(id, username, password, gpgPassphrase)
+  );
+
+  if (gpgPrivateKey !== DEFAULT_GPG_PRIVATE_KEY) {
+    console.log('importing gpg key');
+    const privateKeyDirectory: string = path.join(
+      os.homedir(),
```
clarkbw, line 50: I know this is the common default but in a shared runner I think we wanted to allow for paths outside the home directory such that I could drop this into a temp dir.

__tests__/auth.test.ts
jaredpetersen: I opted for preciseness in the potential future case that someone else decides to call exec. But you're right, there are definitely drawbacks to this approach. I'll change it 👍

src/auth.ts
jaredpetersen: Yeah, I was worried about that. While the private key is being stored in a "temporary" directory inside the home directory, two setup-java actions running on a shared runner could potentially conflict.

Are you suggesting that we allow users to specify a a temporary private key directory, similar to what is already done with the `settings-path` option? Or would you rather see setup-java create a temporary directory using [`os.tmpdir()`](https://nodejs.org/api/os.html#os_os_tmpdir)  based on the [repository + unique run ID](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables) without any prompting. It looks like using `$GITHUB_WORKSPACE/.tmp` or something is an alternative as well.

I was thinking more along the lines of the latter -- creating a known good temporary directory for the user without asking -- since a user shouldn't care about where the private key file is being written because it's being deleted.

Side note -- this comment about shared runners also brings up the potential need for a temporary key ring. After import, the GPG keyring sticks around and can be used by other runs. This wouldn't be a problem with the hosted runner because it's a new environment every time. A better approach for the shared, self-hosted runners might be to [import using a temporary key ring](https://serverfault.com/a/696205) and then clean it up in a [post step](https://help.github.com/en/actions/building-actions/metadata-syntax-for-github-actions#post).

Let me know your thoughts and I can take another crack at this 😄

action.yml
```diff
@@ -36,6 +36,14 @@ inputs:
   settings-path:
     description: 'Path to where the settings.xml file will be written. Default is ~/.m2.'
     required: false
+  gpg-private-key:
+    description: 'Environment variable name for the GPG private key to import. Default is
```
AirQuick, line 40: Is this really an environment variable name? [`README.md`](https://github.com/actions/setup-java/pull/59/files#diff-04c6e90faac2675aa89e2176d2eec7d8R116) appears to suggest it's not. So I'm just wondering.

jaredpetersen: Good catch. It's not an environment variable.

__tests__/auth.test.ts
jaredpetersen: Found another approach that allows a future case where exec is called for something else but also deals with argument drift.

src/auth.ts
joshmgross: For a temporary directory, you can use `RUNNER_TEMP`:

>The path of the temporary directory for the runner. This directory is guaranteed to be empty at the start of each job, even on self-hosted runners.
https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#runner-context

Code example: https://github.com/actions/toolkit/blob/83dd3ef0f1e5bc93c5ab7072e1edf1715a01ba9d/packages/tool-cache/src/tool-cache.ts#L555

jaredpetersen: Ah, thanks! I see now that `installer.ts` is already using this directory as well. Updated PR to use the temp directory but some more improvements can be made since `installer.ts` apparently already has some logic built in for fallbacks. I'll move that common temp directory logic to a separate file that both `auth.ts` and `installer.ts` use in a future commit.

- - -

The temporary key ring thing is unfortunately more complicated than I originally thought. You can't just create and use a temporary key ring without also telling the maven gpg plugin to use the same configuration.

If we want to isolate the gpg keyring for shared runners, the cleanest way I've found is to specify a new GPG home directory that lives in the temp directory. This means that each job will get its own brand new GPG configuration that is cleaned up before the next job run.

To start, we will have to import the private key via:
```
gpg --homedir $RUNNER_TEMP/.gnupg/ --import --batch private.asc
```

However, using `--homedir` only affects that particular import command. Future GPG commands will use the default GPG home directory which is a problem for us when it comes invoking GPG later with Maven. To fix this, we have to provide a home directory override for the GPG plugin too.

The resulting `settings.xml` file would then look like:
```xml
<servers>
    <server>
      <id>github</id>
    </server>
    <server>
      <id>gpg.passphrase</id>
      <passphrase>${env.GPG_PASSPHRASE}</passphrase>
    </server>
    <profiles>
        <profile>
            <activation>
                <activeByDefault>true</activeByDefault>
            </activation>
            <properties>
                <gpg.homedir>${env.RUNNER_TEMP}</gpg.homedir>
            </properties>
        </profile>
    </profiles>
</servers>
```

Creating this default GPG home directory override in the Maven settings **will** break existing setup-java users that have scripted their own GPG import solution. We need to generate a different `settings.xml` for users that haven't specified a private key:

```xml
<servers>
    <server>
      <id>github</id>
    </server>
</servers>
```
_**NOTE:** The second `gpg.passphrase` server section can stay as-is without any harm but we might as well remove it if we're already conditionally generating settings.xml based on parameters._

This is going to be a bit more difficult to implement and it's already late here. I'll get cracking on this when I can.

jaredpetersen:
Still a few kinks to work out it looks like but good progress.

jaredpetersen:
Looks like the homedir approach is a bust as well. Works great everywhere but Windows:

```
"C:\Program Files\Git\usr\bin\gpg.exe" --homedir d:/a/_temp/.gnupg --import --batch private-key.asc
gpg: directory 'd:/a/_temp/.gnupg' created
gpg: keybox 'd:/a/_temp/.gnupg/pubring.kbx' created
gpg: d:/a/_temp/.gnupg/trustdb.gpg: trustdb created
gpg: key F3D9B27AE556B8C0: public key "Jared Petersen <XXXXXXXXX@gmail.com>" imported
gpg: can't connect to the agent: IPC connect call failed
gpg: error getting the KEK: No agent running
gpg: error reading 'private-key.asc': No agent running
gpg: import from 'private-key.asc' failed: No agent running
gpg: Total number processed: 0
gpg:               imported: 1
gpg:       secret keys read: 1
##[error]The process 'C:\Program Files\Git\usr\bin\gpg.exe' failed with exit code 2
```

Trying to start up an agent is not going well either.

I'm going to keep working on this. I think it's valuable for us to clean up the keys we import so that we don't leave anything behind. We'll just have to do it a different way.

Closing this out for now so that I can work on this without pinging everyone constantly.

AirQuick:
```console
C:\Users\MyUser>"%ProgramFiles%\Git\usr\bin\gpg.exe" --version
gpg (GnuPG) 2.2.20-unknown
...
Home: /c/Users/MyUser/.gnupg
```

This seems to suggest dropping `:` and appending `/`, which I tried:

```console
C:\Users\MyUser>tskill gpg-agent
Could not find process: gpg-agent

C:\Users\MyUser>dir c:\Users\MyUser\AppData\Local\Temp\.gnupg /b

C:\Users\MyUser>"%ProgramFiles%\Git\usr\bin\gpg.exe" --homedir "/c/Users/MyUser/AppData/Local/Temp/.gnupg" --import --batch "C:\Users\MyUser\Desktop\gpg-private.key.backup"
gpg: keybox '/c/Users/MyUser/AppData/Local/Temp/.gnupg/pubring.kbx' created
gpg: /c/Users/MyUser/AppData/Local/Temp/.gnupg/trustdb.gpg: trustdb created
gpg: key ...: public key "..." imported
gpg: key ...: secret key imported
gpg: Total number processed: 1                                                                                          gpg:               imported: 1
gpg:       secret keys read: 1
gpg:   secret keys imported: 1

C:\Users\MyUser>dir c:\Users\MyUser\AppData\Local\Temp\.gnupg /b
private-keys-v1.d
pubring.kbx
pubring.kbx~
trustdb.gpg

C:\Users\MyUser>tskill gpg-agent

C:\Users\MyUser>
```

jaredpetersen:
@AirQuick That's a great find. So you didn't have to start up a new GPG agent?

AirQuick:
I didn't have to. The first `tskill gpg-agent` in my example meant to ensure it.

clarkbw:
Also worth noting that this setup-java didn't require it because we stored ENV variables but it is possible to run a cleanup script afterward if you have keys or secrets that should be removed.

Checkout this example:
- https://github.com/clarkbw/setup-docker/blob/master/action.yml#L18
- https://github.com/clarkbw/setup-docker/blob/master/src/cleanup.ts

jaredpetersen:
@clarkbw Yeah, I was thinking of that as an alternative solution since the homedir approach wasn't working very well on Windows. You could use the default homedir to import the key, parse out the key ID from the output, and then remove the key in a cleanup step.

The advantage of that is that approach is that your job can take advantage of other keys that were set up prior on the shared runner. It's probably a healthier approach in the longer term.

Sorry for all the back and forth on this. I use GitHub actions in a hosted-only context so learning about and wrapping my head around the shared paradigm has been a challenge.

PR #63, opened 2020-05-16
Added support for GPG signing

jaredpetersen:
Round 2 for adding support for GPG to setup-java per issue https://github.com/actions/setup-java/issues/43. You can see my previous attempt in PR https://github.com/actions/setup-java/pull/59.

This modifies the generated `settings.xml` file so that [Maven GPG plugin](https://maven.apache.org/plugins/maven-gpg-plugin/) can use it, calls `gpg` to import a private key (using the runner temp directory), and then removes it in a newly-implemented post step. All of the GPG key logic is located in a separate module rather than shoving it into `auth.ts` like my first attempt.

I also added a XML library to make it easier to conditionally generate XML (and has the added benefit of removing the custom XML parsing logic that currently exists). This will also make it easier to do `settings.xml` modifications later if needed.

The generated `settings.xml` file looks the following without GPG enabled:

```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
  <servers>
    <server>
      <id>github</id>
    </server>
  </servers>
</settings>
```

and like the following with GPG enabled:
```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
  <servers>
    <server>
      <id>github</id>
    </server>
    <server>
      <id>gpg.passphrase</id>
      <passphrase>${env.GPG_PASSPHRASE}</passphrase>
    </server>
  </servers>
</settings>
```

You can see some successful cross-platform test runs here:
- [Default, GPG not in use](https://github.com/jaredpetersen/setup-java-gpg/actions/runs/106387705)
- [GPG enabled, default parameters](https://github.com/jaredpetersen/setup-java-gpg/actions/runs/106387701)
- [GPG enabled, override parameters](https://github.com/jaredpetersen/setup-java-gpg/actions/runs/106387703)


jaredpetersen:
CC @joshmgross & @clarkbw

clarkbw:
🤩 cleaned this up a lot! I’ll try to do a review on Tuesday when I’m back.

jaredpetersen:
Great! Enjoy your weekend!

src/auth.ts
```diff
@@ -3,60 +3,72 @@ import * as os from 'os';
...
-  await write(directory, generate(id, username, password));
-}
-
-function escapeXML(value: string) {
-  return value
-    .replace(/&/g, '&amp;')
-    .replace(/</g, '&lt;')
-    .replace(/>/g, '&gt;')
-    .replace(/"/g, '&quot;')
-    .replace(/'/g, '&apos;');
```
clarkbw, line 40: Great to see this function go!

src/setup-java.ts
```diff
@@ -1,10 +1,20 @@
...
+const DEFAULT_ID = 'github';
+const DEFAULT_USERNAME = 'GITHUB_ACTOR';
+const DEFAULT_PASSWORD = 'GITHUB_TOKEN';
+const DEFAULT_GPG_PRIVATE_KEY = undefined;
+const DEFAULT_GPG_PASSPHRASE = 'GPG_PASSPHRASE';
+
 async function run() {
   try {
+    // Set secrets before use
+    core.setSecret('gpg-private-key');
```
clarkbw, line 16: I think you need to do this below and use the `gpgPrivateKey` variable. `setSecret` masks out values you want hidden so this would currently mask the text `gpg-private-key`.

action.yml
```diff
@@ -36,6 +36,14 @@ inputs:
...
+    required: false
+  gpg-passphrase:
+    description: 'Environment variable name for the GPG private key passphrase. Default is
+       $GPG_PASSPHRASE.'
+    required: false
 runs:
   using: 'node12'
-  main: 'dist/index.js'
+  main: 'dist/setup/index.js'
+  post: 'dist/cleanup/index.js'
```
konradpabjan, line 49: My biggest concern right now is the post cleanup. Is this absolutely necessary? 

If users don't use any of the maven functionality and they just want a version of java to be downloaded and set up, they'll still get the cleanup job to show up in their logs: 

image

From the run you linked: https://github.com/jaredpetersen/setup-java-gpg/actions/runs/106387705

In the scenario where a GitHub hosted environment is used, the VM gets destroyed at the very end, so from a security perspective it doesn't make much sense got have to call `gpg.deleteKey`. Is there something I might be missing?

jaredpetersen: The post cleanup is necessary for the shared environment. In the case where people are hosting their own GitHub runners, we don't want want keys kicking around in a GPG keychain after the job completes. They don't have the luxury of a self-cleaning runner.

jaredpetersen: If there's a way to only show that post job cleanup if a user configures GPG, I'd be happy to add it. But as far as I'm aware, the `action.yml` only allows you to run it based on job completion status (like `success`) or at all times.

src/setup-java.ts
jaredpetersen: Good callout. I could have sworn that it wanted the name of the variable or something. I will take another look at the docs and implement this fix as soon as I can!

action.yml
jaredpetersen: I see that there's a post-if doc [here](https://help.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#post-if) but I don't believe I can specify that based on an input.

konradpabjan: So I did some local testing, the `RUNNER_TEMP` directory that is used as part of `getTempDir` gets cleanup up and deleted after each workflow run so it should be safe to store any GPG files. Any of the extra post-cleanup steps could then be avoided.

We have an internal issue open to publicly document `RUNNER_TEMP` and what it actually does, but this seems like it it should do the trick. I'll follow up with some of the folks who are more familiar with `RUNNER_TEMP` and ask if this is safe to use from a security perspective.

konradpabjan: I double checked with @TingluoHuang and the `RUNNER_TEMP` env variable will bet set for both self-hosted and hosted.

The only time the `RUNNER_TEMP` directory will not be cleaned up at the end is if the job doesn't finish (runner crashes or the machine shuts down). The post job cleanup is susceptible to the same thing so it's safe to just omit all this logic (the whole PR can get considerably simplified now 😃 )

Of course, some mention in the `README` would be good so that users are aware of this behavior and where the GPG file gets created. The temp directory is usually `_work/_temp` and it's in the same location as where the self-hosted runner is running from.

jaredpetersen: The GPG private key is written to the temp directory but the GPG keychain is not stored there. The keychain lives in the user's home directory by default so it won't be cleaned up.

I tried to change the GPG home directory in a previous PR but it doesn't work well on Windows because of the file paths. You have to override the directory separators (\) manually to the posix format in order for it to work. The built in path library does not support this behavior so we have to write that ourselves. Not difficult, but dirty.

Overriding the GPG home directory also mandates that you override the configuration in a new active by default profile in the settings.xml file so that the Maven GPG plugin knows where to look.

This whole thing ends up being dirtier than the post cleanup solution and doesn't allow for any ssh keys that a self hosted runner may already have installed for other purposes in the job.

I think importing the key and removing it results in a cleaner settings.xml file and impacts the runner the least. Just import the key and remove it when you're done. It does involve adding a post job, but I think that's a better trade-off.

jaredpetersen: If you both disagree, let me know and I will implement those changes. I just wanted it to be clear what that entails.

konradpabjan: 😨 That sounds pretty awful. The issue with other ssh keys not working also sounds like a nightmare with lots of potential consequences. This makes me lean towards adding the cleanup step at the end. It's not ideal because the majority of users won't need it, but I think it's the lesser of the two evils. 

This post cleanup job could also have some other use cases in the future, so maybe this isn't all that bad. I'm going to ask a few more people internally to see if there are any major objections to this.

jaredpetersen: > It's not ideal because the majority of users won't need it, but I think it's the lesser of the two evils

That's exactly where I'm at at this point. At least the cleanup step is very quick.

> I'm also going to check if there might be some secret way of running the post cleanup only if it's a self-hosted runner

That would be awesome and is something I didn't even think about! If it's a hosted runner, we don't have to clean anything up because the entire runner gets thrown away.

AirQuick: Sorry to rush, but any update?

konradpabjan: Sorry for the delay, but I'm still waiting on an update so nothing to share at this point (I would really like to get this in though).

jaredpetersen:
Merged master into my fork since there were merge conflicts. Let me know if there's anything I can do on my end to push this forward, @konradpabjan 🙂

clarkbw:
With `setSecret` fixed it looks good to me.

jaredpetersen:
@konradpabjan Any updates on your end?

konradpabjan:
All good! No objections @jaredpetersen so we can move forward 😃 

Apologies for the huge wait. I'll do a in-depth review today or tomorrow at the latest and we can get this merged in.

action.yml
```diff
@@ -36,11 +36,19 @@ inputs:
...
+    required: false
 outputs:
   path:
     description: 'Path to where the java environment has been installed (same as $JAVA_HOME)'
   version:
     description: 'Actual version of the java environment that has been installed'
 runs:
   using: 'node12'
-  main: 'dist/index.js'
+  main: 'dist/setup/index.js'
```
konradpabjan, line 53: I like this approach of separate files. 👍

`checkout` has a `post` step as well, but it lumps everything together into a single file. I think what you have is better. https://github.com/actions/checkout/blob/592cf69a223b04e75ddf345919130b91010eb2a6/action.yml#L73

src/cleanup-java.ts
```diff
@@ -0,0 +1,16 @@
+import * as core from '@actions/core';
+import * as gpg from './gpg';
+
+async function run() {
+  if (core.getInput('gpg-private-key', {required: false})) {
+    console.log('removing private key from keychain');
```
konradpabjan, line 6: There is a `core.info` method that can be used over `console.log` https://github.com/actions/toolkit/tree/master/packages/core#logging

Does the same thing effectively, but we prefer to use it.

src/cleanup-java.ts
```diff
@@ -0,0 +1,16 @@
...
+    try {
+      const keyFingerprint = core.getState('gpg-private-key-fingerprint');
```
konradpabjan, line 8: There are a bunch of constants floating around such as `gpg-private-key-fingerprint` and `gpg-private-key`. Could we throw those into a single file such as `constants.ts`?

src/util.ts
```diff
@@ -0,0 +1,26 @@
+import * as path from 'path';
+
+export function getTempDir() {
```
konradpabjan, line 3: I've seen this method copy/pasted a few times throughout our actions 😄  This is fine for now, but I opened an issue in `toolkit` so that it's added to one of our packages: https://github.com/actions/toolkit/issues/518

src/setup-java.ts
```diff
@@ -15,16 +22,32 @@ async function run() {
...
 
-    const matchersPath = path.join(__dirname, '..', '.github');
+    const matchersPath = path.join(__dirname, '..', '..', '.github');
     console.log(`##[add-matcher]${path.join(matchersPath, 'java.json')}`);
 
-    const id = core.getInput('server-id', {required: false}) || undefined;
+    const id = core.getInput('server-id', {required: false}) || DEFAULT_ID;
     const username =
-      core.getInput('server-username', {required: false}) || undefined;
+      core.getInput('server-username', {required: false}) || DEFAULT_USERNAME;
```
konradpabjan, line 30: Given that there are new inputs, this can also go into a constants file. Something like this: https://github.com/actions/download-artifact/blob/main/src/constants.ts

src/setup-java.ts
```diff
@@ -1,8 +1,15 @@
 import * as core from '@actions/core';
 import * as installer from './installer';
 import * as auth from './auth';
+import * as gpg from './gpg';
 import * as path from 'path';
 
+const DEFAULT_ID = 'github';
+const DEFAULT_USERNAME = 'GITHUB_ACTOR';
```
konradpabjan, line 8: Some of these defaults can go straight into `action.yml`

Example: https://github.com/actions/setup-python/blob/654aa00a6ed0d2f7cf168f68bf3cca5daf06cded/action.yml#L13

__tests__/gpg.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const tempDir = path.join(__dirname, 'runner', 'temp');
+process.env['RUNNER_TEMP'] = tempDir;
+
+import gpg = require('../src/gpg');
+
+describe('gpg tests', () => {
+  beforeEach(async () => {
+    await io.mkdirP(tempDir);
+  }, 300000);
```
konradpabjan, line 19: A timeout of 300 seconds seems way too high 🤔 same thing for `afterall`. The default is 5sec so I think it's fine to keep it at that: https://jestjs.io/docs/en/api#beforeallfn-timeout

README.md
```diff
@@ -113,39 +113,58 @@ jobs:
...
     - name: Publish to Apache Maven Central
-      run: mvn deploy 
+      run: mvn deploy
       env:
         MAVEN_USERNAME: maven_username123
         MAVEN_CENTRAL_TOKEN: ${{ secrets.MAVEN_CENTRAL_TOKEN }}
+        MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
 ```
 
 The two `settings.xml` files created from the above example look like the following.
```
konradpabjan, line 127: Some extra info can be added here about the gpg file and the purpose of the post cleanup step.

jaredpetersen:
Sounds good! I will try to address your feedback tonight. The links to examples from other github actions repositories are very much appreciated and will save me a lot of time.

jaredpetersen:
Okay, addressed the feedback. Should be good to go for another round of review.

Did another set of runs on my own repo to verify that everything is still working as well.
- [Default](https://github.com/jaredpetersen/setup-java-gpg/actions/runs/170887431)
- [Default GPG](https://github.com/jaredpetersen/setup-java-gpg/actions/runs/170887433)
- [GPG override](https://github.com/jaredpetersen/setup-java-gpg/actions/runs/170887436)

README.md
```diff
@@ -113,39 +113,60 @@ jobs:
...
+      <passphrase>${env.MAVEN_GPG_PASSPHRASE}</passphrase>
+    </server>
+  </servers>
+</settings>
 ```
 
-***NOTE: The `settings.xml` file is created in the Actions $HOME directory. If you have an existing `settings.xml` file at that location, it will be overwritten. See below for using the `settings-path` to change your `settings.xml` file location.***	
+***NOTE: The `settings.xml` file is created in the Actions $HOME directory. If you have an existing `settings.xml` file at that location, it will be overwritten. See below for using the `settings-path` to change your `settings.xml` file location.***
+
+If `gpg-private-key` and `gpg-passphrase` inputs are provided, the private key will be written to a file in the runner's temp directory, the private key file will be imported into the GPG keychain, and then the file will be promptly removed before proceeding with the rest of the setup process. A cleanup step will remove the imported private key from the GPG keychain after the job completes regardless of the job status. This ensures that the private key is no longer accessible on self-hosted runners and cannot "leak" between jobs (hosted runners are always clean instances).
```
konradpabjan, line 169: 👍

action.yml
```diff
@@ -25,22 +25,33 @@ inputs:
     description: 'ID of the distributionManagement repository in the pom.xml
        file. Default is `github`'
     required: false
+    default: 'github'
   server-username:
     description: 'Environment variable name for the username for authentication
        to the Apache Maven repository. Default is $GITHUB_ACTOR'
     required: false
+    default: 'GITHUB_ACTOR'
```
konradpabjan, line 33: This should be `${{ GITHUB_ACTOR }}`

Env variables need to be wrapped in the special syntax so they correctly get resolved. Example: https://github.com/actions/setup-python/blob/7a69c2bc7dc38832443a11bc7c2550ba96c6f45c/action.yml#L13

action.yml
```diff
@@ -25,22 +25,33 @@ inputs:
...
   server-password:
     description: 'Environment variable name for password or token for
        authentication to the Apache Maven repository. Default is $GITHUB_TOKEN'
     required: false
+    default: 'GITHUB_TOKEN'
```
konradpabjan, line 38: Same as above `${{ GITHUB_TOKEN }}`

__tests__/gpg.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    await io.mkdirP(tempDir);
+  });
+
+  afterAll(async () => {
+    try {
+      await io.rmRF(tempDir);
+    } catch {
+      console.log('Failed to remove test directories');
+    }
+  }, 100000);
```
konradpabjan, line 27: This huge timeout can also be removed

action.yml
jaredpetersen: Are you sure? The defaults look like they were originally the string literals:

https://github.com/actions/setup-java/blob/d920b7da5f18f6c632513f44c068598875961bbb/src/auth.ts#L11-L12

That way they would be written to the settings.xml file with environment variables names rather than the actual environment variable values:

https://github.com/actions/setup-java/blob/d920b7da5f18f6c632513f44c068598875961bbb/src/auth.ts#L54-L55

From the README:

https://github.com/actions/setup-java/blob/master/README.md#publishing-using-apache-maven

```xml
<servers>
    <server>
      <id>github</id>
    </server>
</servers>
```

I believe if I put the wrapping there, the `settings.xml` file would look like:

```xml
<servers>
    <server>
      <id>github</id>
      <username>jaredpetersen</username>
      <password>mysupersecrettokeninplaintext</password>
    </server>
</servers>
```

__tests__/gpg.test.ts
jaredpetersen: Sorry, looks like a bad case of too much copypasta 🍝

action.yml
jaredpetersen: Actually, I think it would look like:

```xml
<servers>
    <server>
      <id>github</id>
      <username>${env.jaredpetersen}</username>
      <password>${env.mysupersecrettokeninplaintext}</password>
    </server>
</servers>
```

jaredpetersen: From my test output, it looks like it's correct as-is. See the `Output settings.xml` step [here](https://github.com/jaredpetersen/setup-java-gpg/runs/876184800?check_suite_focus=true).

jaredpetersen:
Test timeout fixed but I think the `GITHUB_TOKEN` and `GITHUB_ACTOR` feedback deserves another look @konradpabjan.

action.yml
konradpabjan: You're right! I forgot that you can use interpolation and that env variables will be later evaluated.
Something like `${env.HOME}` can be directly in the settings file: https://maven.apache.org/settings.html#quick-overview

src/auth.ts
```diff
@@ -3,60 +3,73 @@ import * as os from 'os';
...
+  password: string,
+  gpgPassphrase: string | undefined = undefined
 ) {
   console.log(
     `creating ${SETTINGS_FILE} with server-id: ${id};`,
-    `environment variables: username=\$${username} and password=\$${password}`
+    'environment variables:',
+    `username=\$${username},`,
+    `password=\$${password},`,
+    `and gpg-passphrase=${gpgPassphrase ? '$' + gpgPassphrase : null}`
```
yacaovsnc, line 23: We should not log these, should we?

README.md
yaananth, line 169:
```suggestion
If `gpg-private-key` input is provided, the private key will be written to a file in the runner's temp directory, the private key file will be imported into the GPG keychain, and then the file will be promptly removed before proceeding with the rest of the setup process. A cleanup step will remove the imported private key from the GPG keychain after the job completes regardless of the job status. This ensures that the private key is no longer accessible on self-hosted runners and cannot "leak" between jobs (hosted runners are always clean instances).
```

gpg-passphrase whether given or not, doesn’t decide how we store the private key

src/cleanup-java.ts
```diff
@@ -0,0 +1,19 @@
...
+async function run() {
+  if (core.getInput(constants.INPUT_GPG_PRIVATE_KEY, {required: false})) {
+    core.info('removing private key from keychain');
+    try {
+      const keyFingerprint = core.getState(
+        constants.STATE_GPG_PRIVATE_KEY_FINGERPRINT
+      );
+      await gpg.deleteKey(keyFingerprint);
+    } catch (error) {
+      core.setFailed(error.message);
```
yaananth, line 14: What could be the error message here? If we fail to clean up, should we surface that as an actionable item to end user so that they can manually clean it up (relevant to self-hosted runners)

src/auth.ts
konradpabjan: This is actually fine, see the output: https://github.com/jaredpetersen/setup-java-gpg/runs/876184775?check_suite_focus=true#step:3:18

In `settings.xml` you can have interpolation: https://maven.apache.org/settings.html#quick-overview with things such as `${env.HOME}` so that nothing gets actually printed to the file. That's what we're doing, just with our own env variables such as $GITHUB_TOKEN

yacaovsnc: If all we are printing out are variable names, then we really don't need to print this right?  It also creates a mental burden for others to check whether we are printing sensitive information or not. 😄 

My vote is to never have anything remotely resembles `print(password)` anywhere.  😃   We could print `password is not null` if that info is needed.

Since this isn't a new problem, this does not block the PR.

yaananth: I think this might help to confirm for example if passphrase is set or not, with `${gpgPassphrase ? '$' + gpgPassphrase : null}` But agreed on the other just `print(pwd)` kinds :)

jaredpetersen: So just to be clear -- we're OK with this current output?

src/cleanup-java.ts
jaredpetersen: This is just going to output whatever we get from running the GPG command. So for example if it failed to find the key for whatever reason, it's going to tell us

```
>> gpg --batch --yes --delete-keys hello
gpg: key "hello" not found: Not found
gpg: hello: delete key failed: Not found
```

Or if it failed, it would just be whatever `@action/exec` surfaces for errors (which may look a little raw).

I can try to simulate what that might look like a little more concretely tonight if I push up a bad cleanup change.

jaredpetersen: It might be safer to just say "failed to delete key" rather than the actual GPG output so that we don't leak anything accidentally. If this fails on a hosted runner, they could run the delete command themselves to get a better understanding of what exactly went wrong.

I think I'll go down that route instead.

action.yml
```diff
@@ -43,6 +47,7 @@ inputs:
     description: 'Environment variable name for the GPG private key passphrase. Default is
        $GPG_PASSPHRASE.'
     required: false
+    default: 'GPG_PASSHPRASE'
```
bissim: I guess `GPG_PASSHPRASE` is a typo.

bissim: nvm, it got removed in some other commit anyway

PR #66, opened 2020-05-25
Add output parameters for the tool path and version

### LINKED ISSUE #65: Add output parameter with path to installed JDK
When calling the action several times, it will overwrite the `JAVA_HOME` environment variable and prepend it to the `PATH`, effectively _shadowing_ a previous run of the action.
Fortunately, it will also set a `JAVA_HOME_<version>_<architecture>` environment variable, but this could be more _idiomatic_ as an output parameter (in addition to the environment variable, which can also be useful on its own: https://github.com/actions/setup-java/issues/44#issuecomment-579776451)

This would be a great addition to #21 

That way, one could write something like:
```yaml
- name: Setup JDK 11
  id: setup-java-11
  uses: actions/setup-java@v1
  with:
    java-version: 11
    # here with #21, one would disable "installation"
- name: Setup JDK 14
  id: setup-java-14
  uses: actions/setup-java@v1
  with:
    java-version: 14
    # here with #21, one would disable "installation"
- name: Setup JDK 8
  # Ran last due each run shadowing the previous ones; see also #21
  uses: actions/setup-java@v1
  with:
    java-version: 8
- name: Build and run tests (with JDK 8)
  run: ./gradlew build
- name: Run tests with JDK 11
  run: ./gradlew test -Ptest.jdk-path="${{ steps.setup-java-11.outputs.jdk-path }}"
- name: Run tests with JDK 14
  run: ./gradlew test -Ptest.jdk-path="${{ steps.setup-java-14.outputs.jdk-path }}"
```
(here, the build is still running with JDK 8, but it coud configure its tests to actually use another JDK)

tbroyer:
This allows calling the action multiple times in the
same job and retrieving the path and/or version in
other steps.

Fixes #65

### DIFF
--- a/__tests__/verify-java.ps1
+++ b/__tests__/verify-java.ps1
@@ -9,3 +9,23 @@ if (!$java_version.Contains($args[0]))
 {
   throw "Unexpected version"
 }
+
+if ($args.Count -lt 2 -or !$args[1])
+{
+  throw "Must supply java path argument"
+}
+
+if ($args[1] -ne $Env:JAVA_HOME)
+{
+  throw "Unexpected path"
+}
+
+if ($args.Count -lt 3 -or !$args[2])
+{
+  throw "Must supply java version argument"
+}
+
+if ($args[0] -ne $args[2])
+{
+  throw "Unexpected version"
+}
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -35,15 +35,16 @@ jobs:
       if: runner.os == 'windows'
       run: move "${{ runner.tool_cache }}" "${{ runner.tool_cache }}.old"
     - name: Setup Java 13
+      id: setup-java
       uses: ./
       with:
         java-version: 13.0.2
     - name: Verify Java 13
       if: runner.os != 'windows'
-      run: __tests__/verify-java.sh 13.0.2
+      run: __tests__/verify-java.sh 13.0.2 "${{ steps.setup-java.outputs.path }}" "${{ steps.setup-java.outputs.version }}"
     - name: Verify Java 13 (Windows)
       if: runner.os == 'windows'
-      run: __tests__/verify-java.ps1 13.0.2
+      run: __tests__/verify-java.ps1 13.0.2 "${{ steps.setup-java.outputs.path }}" "${{ steps.setup-java.outputs.version }}"
 
   test-proxy:
     runs-on: ubuntu-latest
@@ -62,11 +63,12 @@ jobs:
     - name: Clear tool cache
       run: rm -rf $RUNNER_TOOL_CACHE/*
     - name: Setup Java 13
+      id: setup-java
       uses: ./
       with:
         java-version: 13.0.2
     - name: Verify Java 13
-      run: __tests__/verify-java.sh 13.0.2
+      run: __tests__/verify-java.sh 13.0.2 "${{ steps.setup-java.outputs.path }}" "${{ steps.setup-java.outputs.version }}"
 
   test-bypass-proxy:
     runs-on: ubuntu-latest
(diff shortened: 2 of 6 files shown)


__tests__/verify-java.ps1
```diff
@@ -1,11 +1,31 @@
 if (!$args.Count -or !$args[0])
 {
-  throw "Must supply java version argument"
+  throw "::error::Must supply java version argument"
```
konradpabjan, line 3: Looks good! Thanks for submitting this PR.

The only question I have is regarding this `::error::` syntax. Is there any benefit to doing this? I haven't seen any of our other first party actions use this (I want to maintain some level of consistency). Online I can't seen to find anything either.

tbroyer: This is https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message, as seen in `@actions/core`'s `error()` too as an implementation detail.

Not sure about the `throw` in PowerShell though, but according to https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_throw?view=powershell-7 this should work too.

konradpabjan: I played around this, it makes the error message show up red in the logs so it's a pretty nice addition 😃 

image

It doesn't fail the step though so `exit 1` is still needed (for bash).

If Powershell it's a redundant as throw will make the logs show up red and it will fail the step so the extra `::errror::` is not needed

image

konradpabjan, line 3:
```suggestion
  throw "Must supply java version argument"
```

__tests__/verify-java.ps1
```diff
@@ -1,11 +1,31 @@
...
 }
 
 $java_version = & cmd.exe /c "java -version 2>&1" | Out-String
 Write-Host "Found java version: $java_version"
 if (!$java_version.Contains($args[0]))
 {
-  throw "Unexpected version"
+  throw "::error::Unexpected version"
```
konradpabjan, line 10:
```suggestion
  throw "Unexpected version"
```

__tests__/verify-java.ps1
```diff
@@ -1,11 +1,31 @@
...
+}
+
+if ($args.Count -lt 2 -or !$args[1])
+{
+  throw "::error::Must supply java path argument"
```
konradpabjan, line 15:
```suggestion
  throw "Must supply java path argument"
```

__tests__/verify-java.ps1
```diff
@@ -1,11 +1,31 @@
...
+}
+
+if ($args[1] -ne $Env:JAVA_HOME)
+{
+  throw "::error::Unexpected path"
```
konradpabjan, line 20:
```suggestion
  throw "Unexpected path"
```

__tests__/verify-java.ps1
```diff
@@ -1,11 +1,31 @@
...
+}
+
+if ($args.Count -lt 3 -or !$args[2])
+{
+  throw "::error::Must supply java version argument"
```
konradpabjan, line 25:
```suggestion
  throw "Must supply java version argument"
```

__tests__/verify-java.ps1
```diff
@@ -1,11 +1,31 @@
...
+}
+
+if ($args[0] -ne $args[2])
+{
+  throw "::error::Unexpected version"
```
konradpabjan, line 30:
```suggestion
  throw "Unexpected version"
```

tbroyer:
[Force-pushed](https://github.com/actions/setup-java/compare/141e5bf65f7e604724d0596dbe9060a9515cad03..457d7a45797988287acdd155500b99aa2014b79d) with the PowerShell changes.

konradpabjan:
Did some extra testing and everything is 👍

Thanks for the contribution @tbroyer 🎉🎈

Issue #13, opened 2019-08-09
Offer AdoptOpenJDK in addition to Zulu


ppalaga:
Should be downloadable e.g. from https://adoptopenjdk.net/

mattnelson:
OpenJDK should be the default instead a vendor's specific distribution, but there should be an option to opt into a specific vendor(azul, amazon, etc...)

onebeartoe:
I hope I am not hijacking this thread and I'll create a separate issue if requested.

So, I **wget** is used to pull a JDK with JavaFX.

The workflow then calls **ls** and shows the .gz file, but during the setup-java action it says the file specified by **jdkFile** does not exist.

Am I correctly specifing the **jdkFile**?

This is the repository I am trying to use with Github Actions:

https://github.com/onebeartoe/java-libraries/commit/88b3fdc9b76dc6ba366773f75d6683dfc0142ef5/checks

martin-g:
One more vote for OpenJDK being an option!
Apache Wicket project tried GitHub actions but the build failed due to `strange behavior of Java 11`. I've contacted Oracle to ask whether a bug fix from Java 13 has been downported to Java 11 and they explained to me that 11.0.2 is the last release by Oracle. Anything newer is from any of the other vendors and they (Oracle) have no idea what is included in it.
More details about the problem and the discussion can be found at https://markmail.org/message/g5cbnw47kt7yrj7t

damccorm:
Whoops, sorry I missed this, for some reason my alerts for this repo got turned off 😳

I actually originally wanted to do this using OpenJDK, but ran into an issue that I couldn't find a good solution to. Basically the OpenJDK download urls aren't consistent and there's not a well defined distro. So for example, here are the Linux download urls for a some of the versions, pulled from [here](https://jdk.java.net/archive/):

| version | url |
| ------- | --- |
| 12 | https://download.java.net/java/GA/jdk12/33/GPL/openjdk-12_linux-x64_bin.tar.gz
| 11.0.1 | https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linux-x64_bin.tar.gz
| 11 | https://download.java.net/java/ga/jdk11/openjdk-11_linux-x64_bin.tar.gz
| 10.0.2 | https://download.java.net/java/GA/jdk10/10.0.2/19aef61b38124481863b1413dce1855f/13/openjdk-10.0.2_linux-x64_bin.tar.gz
[...]

PR #67, opened 2020-05-26
add support for adoptopenjdk binaries

gdams:
partially resolves https://github.com/actions/setup-java/issues/13


giltene:
The option name “distro” makes much more sense here than “vendor” IMO (see examples at https://github.com/actions/setup-java/issues/13#issuecomment-568232555).

I’d go with “adopt” for the AdoptOpenJDK distro (likely to last longer given recent gripes around the use of the “OpenJDK” trademark in names as opposed to the fair use rules laid out on the OpenJDK site).

I think we should structure to accommodate several distros, and pre-populate some of the additional likely distro names: 

zulu
oracle-openjdk
redhat
corretto
liberica
adopt
dragonwell

This way folks can later just tweak the (initially empty?) implementation for each distro to enable them. I’m pretty sure at least a couple of them will fill those in pretty quickly...

Should we break out the distro-specific items to a separate .ts file?

gdams:
> Should we break out the distro-specific items to a separate .ts file?

@giltene updated to have separate files.

> The option name “distro” makes much more sense here than “vendor” IMO (see examples at #13 (comment)).

Agreed, updated

> I’d go with “adopt” for the AdoptOpenJDK distro (likely to last longer given recent gripes around the use of the “OpenJDK” trademark in names as opposed to the fair use rules laid out on the OpenJDK site).

I'd prefer to stick with AdoptOpenJDK for now. In the event of a rebrand, it will be very easy to resolve this with an alias

giltene:
For the readme, we should probably add a section ( at the bottom?) describing the possible distro options and their limitations (if any). I anticipate that we as we move past having only one distro we will quickly get to 4-7, and with each distro potentially having different limitations (only some version syntax supported, only some package types, platforms, or architectures supported, possibly only some versions or ranges of versions supported), we can state the limitations of each in a conscience way under the concise distro description in the readme.

The version specification description at the top should remain generic (no “Zulu only” or adopt specific notes), and we can add a small subnote saying “see distro-specific limitations below”. A key purpose purpose of the limitation notes (which each distro can list separately if they want) is to intercept a user being surprised that a specification that works in the default case but doesn’t work in a specific distro for some reason (e.g. using semver range syntax fur adopt in current implementation, or “Oracle OpenJDK does not have jdk+fox packages available”). We don’t have to list all limitations, or even any  (you can leave them blank and have people just figure out why something fails), but they would be helpful fir distro to include. As limitations are removed (full semver syntax supported, additional package types or platforms added), distributions can update the notes, but we probably don’t want to be churning the readme often...

gdams:
This should be good to land,  is somebody happy to merge?

giltene:
> This should be good to land, is somebody happy to merge?

Please address the previous notes on the readme changes you propose... (e.g. remove the "XYZ only" additions to some of the version specification syntax, introduce a distros list to the readme, with notes that each distro can have (as needed) about any limitations or specific expectations.)

README.md
```diff
@@ -19,7 +19,8 @@ steps:
 - uses: actions/checkout@v2
 - uses: actions/setup-java@v1
   with:
-    java-version: '9.0.4' # The JDK version to make available on the path.
+    java-version: '11' # The JDK version to make available on the path.
+    vendor: adoptopenjdk # (adoptopenjdk or zulu) - defaults to adoptopenjdk
```
maxim-lobanov, line 23: Looks like it should be `distro` instead of `vendor`?

package.json
```diff
@@ -6,8 +6,8 @@
   "main": "dist/index.js",
   "scripts": {
     "build": "ncc build -o dist/setup src/setup-java.ts && ncc build -o dist/cleanup src/cleanup-java.ts",
-    "format": "prettier --write **/*.ts",
-    "format-check": "prettier --check **/*.ts",
+    "format": "prettier --write **/*.ts src/distro/*.ts",
```
maxim-lobanov, line 9: nitpick: I think `src/distro/*.ts` should be covered by ` **/*.ts`?

src/installer.ts
```diff
@@ -23,31 +26,26 @@ export async function getJava(
...
+          case 'adoptopenjdk':
+            [jdkFile, version] = await getJavaAdoptOpenJDK(
+              version,
+              javaPackage,
+              arch
+            );
+            break;
+          case 'zulu':
+            [jdkFile, version] = await getJavaZulu(version, javaPackage);
+            break;
```
maxim-lobanov, line 42: I suggest to move this function outside of `getJava` function to improve readability

src/distro/zulu.ts
```diff
@@ -0,0 +1,150 @@
...
+import * as semver from 'semver';
+
+const IS_WINDOWS = process.platform === 'win32';
+
+export async function getJavaZulu(version: string, javaPackage: string) {
+  core.debug('Downloading JDK from Azul');
+  const http = new httpm.HttpClient('setup-java', undefined, {
+    allowRetries: true,
+    maxRetries: 3
+  });
```
maxim-lobanov, line 13: May be we could use some inheritance here.
We can create base class `JavaInstaller` and two child classes: `AzulInstaller`, `AdoptInstaller`. It would allow us to move some common logic to base class and avoid code duplication.
For example, http initialization could be put to constructor and re-used.

src/installer.ts
giltene: I'd also echo the "change `vendor` to `distro`" and ask to use that everywhere for consistency,  including in the code. So this should be setupDistro()...

package.json
gdams: that was what I thought but for some reason prettier wasn't picking it up

konradpabjan:
We've decided to move forward with a `v2` version of `setup-java` that will add support for adoptopenJDK binaries and default to them. I've created a PR with an ADR (architecture decision record), that roughly outlines and documents the roadmap and plans for the v2. Because the default version is changing, a major version update is warranted (v2): https://github.com/actions/setup-java/pull/97

A lot of the hard work has already been done in this PR so we can use this once the ADR is accepted. A bit of restructuring will be needed but we'll get this across the finish line 😃 

If anyone would like to offer some feedback or suggestions in the ADR, I would greatly appreciate it!

cc @giltene @gdams

maxim-lobanov:
Closing since V2 was released recently.

PR #78, opened 2020-06-17
Normalize extendedJavaHome environment variable

aibaars:
The extendedJavaHome environment variable contains `.` symbols in the version.  ~~This causes the environment variable to be ignored by the action runner.~~  This makes the variable inaccessible with `bash`'s `${VAR}` syntax . It's best to replace those and other non standard symbols with and underscore. 

For reference:
> Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '_' (underscore) from the characters defined in Portable Character Set .

See also: https://unix.stackexchange.com/questions/93532/exporting-a-variable-with-dot-in-it

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -87,6 +87,7 @@ export async function getJava(
   }
 
   let extendedJavaHome = 'JAVA_HOME_' + version + '_' + arch;
+  core.exportVariable(extendedJavaHome, toolPath); //TODO: remove for v2
   // For portability reasons environment variables should only consist of
   // uppercase letters, digits, and the underscore. Therefore we convert
   // the extendedJavaHome variable to upper case and replace '.' symbols and
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -87,6 +87,7 @@ export async function getJava(
   }
 
   let extendedJavaHome = 'JAVA_HOME_' + version + '_' + arch;
+  extendedJavaHome = extendedJavaHome.toUpperCase().replace(/[^0-9A-Z_]/g, '_');
   core.exportVariable('JAVA_HOME', toolPath);
   core.exportVariable(extendedJavaHome, toolPath);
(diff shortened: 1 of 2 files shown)


src/installer.ts
```diff
@@ -87,6 +87,7 @@ export async function getJava(
   }
 
   let extendedJavaHome = 'JAVA_HOME_' + version + '_' + arch;
+  extendedJavaHome = extendedJavaHome.toUpperCase().replace(/[^0-9A-Z_]/g, '_');
```
konradpabjan, line 90: My only suggestion would be to add a comment here as to why this is being done. Better to have some comment/link in the code vs hunting down a PR later

aibaars: Makes sense. Done!

src/installer.ts
```diff
@@ -87,6 +87,11 @@ export async function getJava(
   }
 
   let extendedJavaHome = 'JAVA_HOME_' + version + '_' + arch;
+  // For portability reasons environment variables should only consist of
+  // uppercase letters, digits, and the underscore. Therefore we convert
+  // the extendedJavaHome variable to upper case and replace '.' symbols and
+  // any other non-alphanumeric characters with an underscore.
+  extendedJavaHome = extendedJavaHome.toUpperCase().replace(/[^0-9A-Z_]/g, '_');
```
ericsciple, line 94: This looks like a breaking change. Should the old variable be set also, so existing customers dont get broken?

konradpabjan: Good point. It could break certain users if they're actually able to use the current variable.

When we eventually release a v2 version of setup-java, then we can remove the old one (we don't promise compat between the new versions).

aibaars: Sounds good, I just pushed a new commit that keeps the old variable.

PR #82, opened 2020-07-20
Add `overwrite-settings` to disable overwriting settings.xml

### LINKED ISSUE #79: Provide an option to by pass generating a settings.xml file
When using self-hosted runners we provision compute with a Maven `settings.xml` and currently this action will overwrite our correct `settings.xml` with one that's inoperable. It would be nice to signal to the action to not generate a `settings.xml` file.

Another option might be to allow a template `.github/workflow/maven/settings.xml` that can be provisioned on the runner. In the case of running with GitHub actions proper the template can use `${{ secrets.references }}` and in self-host just normal envars which Maven will interpolate into the settings file. Either way it's secure.

joschi:
In some scenarios there's already a valid Maven settings file which shouldn't or cannot be overwritten by actions/setup-java.

The `overwrite-settings` (default: `true`) configuration setting enables users to disable the generation of the Maven settings file by actions/setup-java.

Closes #79
Refs joschi/setup-jdk#11

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -36,6 +36,9 @@ inputs:
   settings-path:
     description: 'Path to where the settings.xml file will be written. Default is ~/.m2.'
     required: false
+  overwrite-settings:
+    description: 'Overwrite the settings.xml file if it exists. Default is "true".'
+    required: false
 outputs:
   path:
     description: 'Path to where the java environment has been installed (same as $JAVA_HOME)'
--- a/src/setup-java.ts
+++ b/src/setup-java.ts
@@ -23,8 +23,15 @@ async function run() {
       core.getInput('server-username', {required: false}) || undefined;
     const password =
       core.getInput('server-password', {required: false}) || undefined;
+    const overwriteSettings =
+      core.getInput('overwrite-settings', {required: false}) || 'true';
 
-    await auth.configAuthentication(id, username, password);
+    await auth.configAuthentication(
+      id,
+      username,
+      password,
+      overwriteSettings === 'true'
+    );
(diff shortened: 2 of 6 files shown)


davidkarlsen:
Anything holding this?

joschi:
@clarkbw @konradpabjan Is there any interested in me updating this PR to the latest `main` branch or should we close this PR?

action.yml
```diff
@@ -36,6 +36,9 @@ inputs:
   settings-path:
     description: 'Path to where the settings.xml file will be written. Default is ~/.m2.'
     required: false
+  overwrite-settings:
+    description: 'Overwrite the settings.xml file if it exists. Default is "true".'
+    required: false
```
maxim-lobanov, line 41: I suggest setting `default: true` in YAML file.
Details: https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputsinput_iddefault

src/setup-java.ts
```diff
@@ -23,8 +23,15 @@ async function run() {
       core.getInput('server-username', {required: false}) || undefined;
     const password =
       core.getInput('server-password', {required: false}) || undefined;
+    const overwriteSettings =
```
maxim-lobanov, line 26: I suggest to use the same approach like `actions/checkout`: https://github.com/actions/checkout/blob/main/src/input-helper.ts#L118
May be we can wrap it in small util func

maxim-lobanov:
Closing as duplicate of https://github.com/actions/setup-java/pull/136

PR #95, opened 2020-08-20
Architecture Support

### LINKED ISSUE #55: `architecture` parameter is ignored

AustinShalit:
Closes #55 

Supersedes #57 and responds to comments in that PR.


AustinShalit:
@joshmgross does this address your comments on #57?

konradpabjan (approved):
Looks good! Thanks for the PR ❤️ 

I noticed that we don't need `dist/index.js` anymore since we have separate pre and post steps: https://github.com/actions/setup-java/blob/3019d15cad3f7d2657e77aa6efe0571b3a89d0b4/action.yml#L56-L57

The entire file should be safety deleted now. I'll go ahead and take care of removing that before releasing the next version of `setup-java`.

PR #96, opened 2020-08-24
Remove old dist/index.js file

konradpabjan:
After [adding support for publishing packages with a GPG key](https://github.com/actions/setup-java/pull/63), there is no longer a single `dist/index.js` file. Instead there is `dist/setup/index.js` and `dist/cleanup/index.js` because there is a main and post step: https://github.com/actions/setup-java/blob/d34a7e45c8812086e1d752fd33b1ba46216a6fdd/action.yml#L56-L57

This PR removes the old `dist/index.js` file that was not removed and it adds some clarification to the contributor docs about the files that `@zeit/ncc` creates.


anthonysterling (approved):
Lovely, thanks @konradpabjan 🙇‍♂️

Issue #72, opened 2020-05-28
Add support for SapMachine binaries


giltene:
With work being done towards adding additional binary distributions of OpenJDK as options (likely under the "distro" option, see e.g. #67), it would be useful to include the SapMachine binaries among the distro options

anilkreddy:
Is this planned for any upcoming releases to include SapMachine binaries?

uluzox:
I would highly appreciate if Sap Machine JVM would be added to `actions/setup-java`

skateball:
Hi all,
it is merged and [released](https://github.com/actions/setup-java/releases/tag/v4.3.0) some minutes ago.
Have fun
Christian

Issue #71, opened 2020-05-28
Add support for Dragonwell binaries


giltene:
With work being done towards adding additional binary distributions of OpenJDK as options (likely under the "distro" option, see e.g. #67), it would be useful to include the dragonwell binaries among the distro options

WtfJoke:
The links can be fetched here: https://dragonwell-jdk.io/releases.json

Issue #70, opened 2020-05-28
Add support for Liberica binaries


giltene:
With work being done towards adding additional binary distributions of OpenJDK as options (likely under the "distro" option, see e.g. https://github.com/actions/setup-java/pull/67), it would be useful to include the Liberica binaries among the distro options

tomaszmichalak:
+1, by default, the Paketo Java buildpack use the Liberica JVM - this buildpack is used by Spring Boot.

rafaelargenta-quanto:
+1, by default, the Paketo Java buildpack use the Liberica JVM - this buildpack is used by Spring Boot.

dmitry-shibanov:
Hello everyone. [Support for Liberica JDK](https://github.com/actions/setup-java/pull/225) was added. I'm going to close the issue.
 - For details see this [Release](https://github.com/actions/setup-java/releases/tag/v2.4.0)

Issue #69, opened 2020-05-28
Add support for Oracle OpenJDK binaries


giltene:
With work being done towards adding additional binary distributions of OpenJDK as options (likely under the "distro" option, see e.g. https://github.com/actions/setup-java/pull/67), it would be useful to include the Oracle OpenJDK binaries among the distro options

sormuras:
In the meanwhile, I published https://github.com/sormuras/download-jdk -- which downloads a JDK from https://jdk.java.net that is consumable by `actions/setup-java` as two-step installation routine.

sormuras:
FYI, I updated [sormuras/download-jdk](https://github.com/marketplace/actions/download-jdk)`@v1` to version 1.1:

- Supports Linux, Mac OS, and Windows
- Supports Early-Access builds of top-level projects, like Loom, Panama, or Valhalla
- Supports archived (EOL!) builds

rahim873:
Google
andriod

Pada Isn, 12 Okt 2020 14:25 Christian Stein <notifications@github.com>
menulis:

xor-freenet:
Good reason to fix this:
Consider how large a discussion about which of the non-Oracle JDKs to use can grow: https://news.ycombinator.com/item?id=28820601
Not everybody has the time to dive into that much complexity just to run their code :)

xor-freenet:
@carimura 
Thanks for the effort! :)

Actions can have very high level of access to a repository and thus a high security impact.  
I would therefore be reluctant to run any Actions which are not part of the official ones of GitHub.

Thus please do indeed ship your code as a PR to this official repository :)

berezovskyi:
@xor-freenet given that they rewrote all the TS code in Java from scratch, it may be a bit challenging...

slawekjaranowski:
I'm also for merging `oracle-actions/setup-java` into `actions/setup-java`

With two actions using matrix for jdk is difficult ....

delabassee:
@slawekjaranowski @berezovskyi https://inside.java/2022/03/11/setup-java/ explains why we needed a separate action

berezovskyi:
@delabassee I buy all the arguments except why a rewrite from TypeScript was needed. TypeScript is not my favourite language but I think all of the changes Oracle team needed could have been implemented as a PR. Anyway, I am awaiting native support for #68 more eagerly as NFTC only vaguely permits "running the Program for Your own personal use or _internal_ business operations".

fniephaus:
[...]

PR #97, opened 2020-08-24
V2 setup-java ADR

konradpabjan:
## Overview

This PR outlines the basic roadmap & plan for the `v2` version of `setup-java`. 

🖌 [v2 ADR rendered](https://github.com/actions/setup-java/blob/v2-adr/docs/adrs/0000-v2-setup-java.md) 🎨

In order to support https://github.com/actions/setup-java/issues/13, the action will have to be overhauled pretty extensively to support downloading different distributions of java. The same work can then be used to support even more distributions in the future.

- https://github.com/actions/setup-java/issues/68 (Amazon Corretto)
- https://github.com/actions/setup-java/issues/69 (Oracle OpenJDK)
- https://github.com/actions/setup-java/issues/70 (Liberica)
- https://github.com/actions/setup-java/issues/71 (Dragonwell)
- https://github.com/actions/setup-java/issues/72 (SapMachine)

As part of `v2`, the default downloaded version will switch to `adoptOpenJDK` instead of `Zulu`. This is a major breaking change that warrants a major version upgrade (`v1` -> `v2`).

Some existing work was done by users to support `adoptOpenJDK` (https://github.com/actions/setup-java/pull/67). That work can be leveraged after a consensus is reached with this ADR.

## We need your help! 

Any ideas, suggestions, or feedback will be invaluable during planing ❤️

### DIFF
--- a/docs/adrs/0000-v2-setup-java.md
+++ b/docs/adrs/0000-v2-setup-java.md
@@ -1,0 +1,74 @@
+# 0. V2 setup-java
+
+Date: 2020-08-24
+
+Status: Proposed
+
+# Context
+
+- The `v1` version of `setup-java` downloads and installs Zulu builds of OpenJDK. There is a huge ask from customers to offer AdoptOpenJDK builds of OpenJDK: https://github.com/actions/setup-java/issues/13
+
+- Zulu and AdoptOpenJDK aren't the only distributions of Java though. Other providers include Oracle OpenJDK or Amazon Corretto and ideally it would be nice to support downloading Java from all providers.
+
+- GitHub Actions virtual environments install and default to AdoptOpenJDK builds of OpenJDK. `setup-java` should align to to the same behavior.
+
+# Decision
+
+## New input
+
+A new input parameter (titled `distribution`) will be added to `setup-java` that will allow users to specify the distribution that they would like to download
+
+```yaml
+  with:
+    java-version: '11'
+    distribution: adoptopenjdk
+```
+
+## Default Behavior
+
+If no `distribution` parameter is provided, the default value will be `adoptopenjdk`. The action will first attempt to use pre-installed versions of AdoptOpenJDK builds of OpenJDK on GitHub hosted runners. If a specific version is not found, the action will then download and install the correct version.
+
+## Extensibility & Documentation
+
+`setup-java` should be structured in such a way that will allow the open source community to easily add support for extra distributions.
+
+Existing code will be restructured so that distribution specific code will be easily separated. Currently the core download logic is in a single file, `installer.ts`. This file will be split up and abstracted out so that there will be no vendor specified logic. Each distribution will have it's own file under `src/distributions` that will contain the core setup logic for a specific distribution. 
+
+```yaml
+ ∟ src/
+    installer.ts # core installer logic
+    distributions/
+        ∟ adoptOpenJDK.ts # adoptOpenJDK specific logic
+        ∟ zulu.ts # zulu specific logic
+        ∟ # future distributions will be added here 
+```
+
+Each specific distribution can have minute differences or other information that should be documented. There will be a directory for distribution specific information under `docs/`. The main `README.md` will document all available distributions and link to specific documentation. The main `README.md` should not contain distribution specific information.
+
+```yaml
+ ∟ docs/
+    distributions/    
+        ∟ adoptOpenJDK.md # adoptOpenJDK specific info
+        ∟ zulu.md # zulu specific info
+        ∟ # future distributions will be added here 
+
+```
+
+The contribution doc (`CONTRIBUTING.md`) will describe how a new distribution should be added and how everything should be structured.
+
+## v2-preview
+
+There will be a `v2-preview` branch that will be created for development and testing. Any changes will first be merged into `v2-preview` branch. After a period of testing & verification, the `v2-preview` branch will be merged into the `main` branch and a `v2` tag will be created. Any [GitHub public documentation](https://docs.github.com/en/actions/language-and-framework-guides/github-actions-for-java) and [starter workflows](https://github.com/actions/starter-workflows) that mention `setup-java` will then be updated to use `v2` instead of `v1`.
+
+## Goals & Anti-Goals
+
+The main focus of the `v2` version of `setup-java` will be to add support for adoptOpenJDK builds of openJDK in addition to Zulu builds of openJDK. In addition, extensibility will be a priority so that other distributions can be added in the future.
+
+The `setup-java` action has some logic that creates a `settings.xml` file so that it is easier to publish packages. Any improvements or modifications to this logic or anything Gradle/Maven specific will be avoided during the development of the `v2-preview`.
+
+# Consequences
+
+- Users will have more flexibility and the freedom to choose a specific distribution that they would like (AdoptOpenJDK builds of OpenJDK in addition or Zulu builds of OpenJDK)
+- `setup-java` will be structured in such a way that will allow for more distributions to be easily added in the future
+- A large subset of users pin to `@main` or `@master` instead of to a specific version (`v1`). By introducing a breaking change by switching from `Zulu` to `AdoptOpenJDK` as the default, certain existing workflows for users might break
+- Higher maintenance and support burden moving forward
--- a/docs/adrs/0000-v2-setup-java.md
+++ b/docs/adrs/0000-v2-setup-java.md
@@ -6,11 +6,11 @@ Status: Proposed
 
 # Context
 
-- The `v1` version of `setup-java` downloads and installs Zulu builds of OpenJDK. There is a huge ask from customers to offer AdoptOpenJDK builds of OpenJDK: https://github.com/actions/setup-java/issues/13
+- The `v1` version of `setup-java` downloads and installs Zulu builds of OpenJDK. There is a huge ask from customers to offer AdoptOpenJDK/Adoptium builds of OpenJDK: https://github.com/actions/setup-java/issues/13
 
 - Zulu and AdoptOpenJDK aren't the only distributions of Java though. Other providers include Oracle OpenJDK or Amazon Corretto and ideally it would be nice to support downloading Java from all providers.
 
-- GitHub Actions virtual environments install and default to AdoptOpenJDK builds of OpenJDK. `setup-java` should give users an option to donwload and use other distributions that may not be be installed
+- GitHub Actions virtual environments install and default to AdoptOpenJDK builds of OpenJDK. `setup-java` should give users an option to download and use other distributions that may not be be installed
 
 # Decision
 
@@ -21,39 +21,28 @@ A new required input parameter (titled `distribution`) will be added to `setup-j
 ```yaml
   with:
     java-version: '11'
-    distribution: adoptopenjdk
+    distribution: adoptium
 ```
 
 ## Default Behavior
 
 There will be no default distribution that we pick for the user. Users will have to specify a distribution in their YAML or else the action will fail.
 
-Requiring a default version will break users that are pinned to `@master` or `@main` as they will have no `distribution` specified in their YAML. Users pinned to `v1` will be uneffected. This change is meant to not be backward compatible and it is acceptable to change the default behavior because a new major version will be released alongside these changes.
+Requiring a default version will break users that are pinned to `@main` as they will have no `distribution` specified in their YAML. Telemetry indicates that only a small percentage of users would be effected though. Users pinned to `v1` will be uneffected. This change is meant to not be backward compatible and it is acceptable to change the default behavior because a new major version will be released alongside these changes.
 
 ## Extensibility & Documentation
 
 `setup-java` should be structured in such a way that will allow the open source community to easily add support for extra distributions.
 
-Existing code will be restructured so that distribution specific code will be easily separated. Currently the core download logic is in a single file, `installer.ts`. This file will be split up and abstracted out so that there will be no vendor specified logic. Each distribution will have it's own file under `src/distributions` that will contain the core setup logic for a specific distribution. 
+Existing code will be restructured so that distribution specific code will be easily separated. Currently the core download logic is in a single file, `installer.ts`. This file will be split up and abstracted out so that there will be no vendor specified logic. Each distribution will have it's own files under `src/distributions` that will contain the core setup logic for a specific distribution. 
 
 ```yaml
  ∟ src/
     installer.ts # core installer logic
     distributions/
-        ∟ adoptOpenJDK.ts # adoptOpenJDK specific logic
-        ∟ zulu.ts # zulu specific logic
+        ∟ adoptium/ # adoptium (formerly AdoptOpenJDK) specific logic
+        ∟ zulu/ # zulu specific logic
         ∟ # future distributions will be added here 
-```
-
-Each specific distribution can have minute differences or other information that should be documented. There will be a directory for distribution specific information under `docs/`. The main `README.md` will document all available distributions and link to specific documentation. The main `README.md` should not contain distribution specific information.
-
-```yaml
- ∟ docs/
-    distributions/    
-        ∟ adoptOpenJDK.md # adoptOpenJDK specific info
-        ∟ zulu.md # zulu specific info
-        ∟ # future distributions will be added here 
-
 ```
 
 The contribution doc (`CONTRIBUTING.md`) will describe how a new distribution should be added and how everything should be structured.
@@ -64,7 +53,7 @@ There will be a `v2-preview` branch that will be created for development and tes
 
 ## Goals & Anti-Goals
 
-The main focus of the `v2` version of `setup-java` will be to add support for adoptOpenJDK builds of openJDK in addition to Zulu builds of openJDK. In addition, extensibility will be a priority so that other distributions can be added in the future.
+The main focus of the `v2` version of `setup-java` will be to add support for adoptium builds of openJDK in addition to Zulu builds of openJDK. In addition, extensibility will be a priority so that other distributions can be added in the future.
 
 The `setup-java` action has some logic that creates a `settings.xml` file so that it is easier to publish packages. Any improvements or modifications to this logic or anything Gradle/Maven specific will be avoided during the development of the `v2-preview`.
 
--- a/docs/adrs/0000-v2-setup-java.md
+++ b/docs/adrs/0000-v2-setup-java.md
@@ -10,7 +10,7 @@ Status: Proposed
 
 - Zulu and AdoptOpenJDK aren't the only distributions of Java though. Other providers include Oracle OpenJDK or Amazon Corretto and ideally it would be nice to support downloading Java from all providers.
 
-- GitHub Actions virtual environments install and default to AdoptOpenJDK builds of OpenJDK. `setup-java` should align to to the same behavior.
+- GitHub Actions virtual environments install and default to AdoptOpenJDK builds of OpenJDK. `setup-java` should align to the same behavior.
 
 # Decision
 
--- a/docs/adrs/0000-v2-setup-java.md
+++ b/docs/adrs/0000-v2-setup-java.md
@@ -10,13 +10,13 @@ Status: Proposed
 
 - Zulu and AdoptOpenJDK aren't the only distributions of Java though. Other providers include Oracle OpenJDK or Amazon Corretto and ideally it would be nice to support downloading Java from all providers.
 
-- GitHub Actions virtual environments install and default to AdoptOpenJDK builds of OpenJDK. `setup-java` should align to the same behavior.
+- GitHub Actions virtual environments install and default to AdoptOpenJDK builds of OpenJDK. `setup-java` should give users an option to donwload and use other distributions that may not be be installed
 
 # Decision
 
 ## New input
 
-A new input parameter (titled `distribution`) will be added to `setup-java` that will allow users to specify the distribution that they would like to download
+A new required input parameter (titled `distribution`) will be added to `setup-java` that will allow users to specify the distribution that they would like to download
 
 ```yaml
   with:
@@ -26,7 +26,9 @@ A new input parameter (titled `distribution`) will be added to `setup-java` that
 
 ## Default Behavior
 
-If no `distribution` parameter is provided, the default value will be `adoptopenjdk`. The action will first attempt to use pre-installed versions of AdoptOpenJDK builds of OpenJDK on GitHub hosted runners. If a specific version is not found, the action will then download and install the correct version.
+There will be no default distribution that we pick for the user. Users will have to specify a distribution in their YAML or else the action will fail.
+
+Requiring a default version will break users that are pinned to `@master` or `@main` as they will have no `distribution` specified in their YAML. Users pinned to `v1` will be uneffected. This change is meant to not be backward compatible and it is acceptable to change the default behavior because a new major version will be released alongside these changes.
 
 ## Extensibility & Documentation
 
@@ -70,5 +72,5 @@ The `setup-java` action has some logic that creates a `settings.xml` file so tha
 
 - Users will have more flexibility and the freedom to choose a specific distribution that they would like (AdoptOpenJDK builds of OpenJDK in addition or Zulu builds of OpenJDK)
 - `setup-java` will be structured in such a way that will allow for more distributions to be easily added in the future
-- A large subset of users pin to `@main` or `@master` instead of to a specific version (`v1`). By introducing a breaking change by switching from `Zulu` to `AdoptOpenJDK` as the default, certain existing workflows for users might break
+- A large subset of users pin to `@main` or `@master` instead of to a specific version (`v1`). By introducing a breaking change that now requires a distribution to be specified, many users will have their workflows fail until they go and update their yaml
 - Higher maintenance and support burden moving forward
--- a/docs/adrs/README.md
+++ b/docs/adrs/README.md
@@ -2,7 +2,7 @@
 
 ADR, short for "Architecture Decision Record" is a way of capturing important architectural decisions, along with their context and consequences.
 
-This folder includes ADRs for the Ecosystem Events team. ADRs are proposed in the form of a pull request, and they commonly follow this format:
+This folder includes ADRs for the setup-java action. ADRs are proposed in the form of a pull request, and they commonly follow this format:
 
 * **Title**: short present tense imperative phrase, less than 50 characters, like a git commit message.
 
--- a/docs/adrs/README.md
+++ b/docs/adrs/README.md
@@ -1,0 +1,19 @@
+# ADRs
+
+ADR, short for "Architecture Decision Record" is a way of capturing important architectural decisions, along with their context and consequences.
+
+This folder includes ADRs for the Ecosystem Events team. ADRs are proposed in the form of a pull request, and they commonly follow this format:
+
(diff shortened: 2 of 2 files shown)


maxim-lobanov:
@konradpabjan , probably, my point is more technical detail and outside of this PR but I think we need to define how `setup-java` will cache versions locally and retrieve from cache:
- Current implementation of `setup-java` use the following tool names in `tc.find(<toolname>, version)`: `jre`, `jdk`, `jdk+fx`. I think the new version of task should include distribution name to tool name to make sure that different distributions of Java cached properly on machines and don't mix up.
- Currently, virtual-environments images don't follow tool-cache layout for Java since we use package managers to install AzulOpenJDK (apt, brew)

brunoborges:
@maxim-lobanov  I'd suggest to avoid `jdk+fx` and `jre` and stick to `jdk` binaries only.

`jdk+fx`: JavaFX / OpenJFX is no longer part of OpenJDK and distributions that provide FX bundled within are basically adding bits. By allowing this, we'd be opening up for variants such as `jdk+swt` or `jdk+netbeansplatform` and so on.

`jre`: starting with Java 9+, there is no longer the concept of JRE, despite some distributions still offering a "JRE" binary that is basically a strimmed down binary of the JDK. Maintaining support of `jre` would perpetuate the confusion. Besides, I really don't see a solid reason to why someone would really want to use a JRE instead of the JDK for a CICD build. Download times are insignificant when building in the cloud, and major JDKs are likely to be either pre-installed or cached.

This action though may provide an option that the user may indicate a remote URL pointing to a very specific/custom binary. Perhaps even one the user built by itself for whatever reason. This argument could then be leveraged for those who want something like a `jdk+fx` or a `jre`.

giltene:
A key comment/question: what is the motivation and justification behind the switch of the default from Zulu to AdoptOpenJDK? An argument for having no default would be one thing (i.e. force a choice of a distro and refuse to work without one), but switching the default from one free distro to another seems like a choice that would require at least some significant amount of open debate before taking action...

I'd highlight some considerations that IMO stand out as important for a default (as long as one exists):

- Java version spectrum: the current default supports (to of the box, with no other spelling needed) java versions 6,7,8,9,10,11,12,13,14, as well as 15-ea and 16-ea, and is clearly likely to continue doing so with additional versions in a timely manner. 

- Versions and continued updates: The current default disto continues to update all active OpenJDK versions as quarterly updates come out in the upstream. E.g. 7u and 13u are currently-active OpenJDK projects with ongoing quarterly updates. (13.0.4 is the current OpenJDK 13u version, 7u272 is the current OpenJDK 7u version, and both are up to date in the current default). Other distros may start doing the same (and hopefully will) but currently don't. Plans to do the same with the upstream 15u (e.g. with OpenJDK 15u 15.0.3 planned for April 2021) are pretty clear as well. There are quite a few projects out there that test against more than just 8 and 11. and projects supporting of 7u, 13u, and 15u are quite likely to want to follow and automatically test on the latest versions of those, in addition to known fixed update levels.

- Continued (no-workflow-change-needed) availability of stale, no-longer-being updated versions: Even when some OpenJDK versions no longer get updated (e.g. 6u, 9u, 12u, 14u are all currently in that state), versions that existed in the past remain available without needing changes to workflow text [This is in contrast to e.g. the situation with other sources for e.g. 9, 10, 12, where stable Travis CI scripts that relied on them broke en-mass when archiving practices and locations changed in the past).

- Support for popularly requested package types (beyond a base JDK): Both jdk+fx and jre packages exist for a reason, and both are in popular demand. They are both commonly used by various projects. See e.g. https://github.com/actions/setup-java/issues/18 for example and for some discussion about the logic for jdk+fx specifically (and why it is much more popular than one might initially think). The jre stuff is often driven by container image size motivations. While some distros include support for these packages and some do not, the majority of OpenJDK distros deliver at least one of these (with equivalents for either "jre" or "jdk+fx" or for both).  E.g. Zulu, Corertto, and Liberica are all examples of distros that have packages that include FX. And Zulu, AdoptOpenJDK, Liberica, Corretto, Liberica are all examples of distros that have packages that are either "jre" or "jre equivalent" (including for e.g. 11u).  

I'm looking forward to seeing some open discussion of counter-arguments and reasons for switching away from the current default.

brunoborges:
> Support for popularly requested package types (beyond a base JDK): Both jdk+fx and jre packages exist for a reason, and both are in popular demand. They are both commonly used by various projects. See e.g. #18 for example and for some discussion about the logic for jdk+fx specifically (and why it is much more popular than one might initially think). The jre stuff is often driven by container image size motivations. While some distros include support for these packages and some do not, the majority of OpenJDK distros deliver at least one of these (with equivalents for either "jre" or "jdk+fx" or for both). E.g. Zulu, Corertto, and Liberica are all examples of distros that have packages that include FX. And Zulu, AdoptOpenJDK, Liberica, Corretto, Liberica are all examples of distros that have packages that are either "jre" or "jre equivalent" (including for e.g. 11u).

Hi Gil,

Let me address this one first as it directly touches my comment. 

This action's goal is to set up Java - the JDK - for the purpose of a CICD build environment in GitHub Actions for general purpose Java applications. For this reason I recommend we keep it simple.

Given that not all distributions provide a bundle with JavaFX, it would be inconvenient to have this action listing X vendors, but only X-N  offering a particular package. For that reason, I support keeping `jdk+fx` out. Users and vendors are still welcome to create their own GH Actions and publish on the Markteplace. The OpenJFX team may as well create their own Action to centralize all JDK+JavaFX bundles in a single action, and then users could use that instead of Setup Java.

Similar situation happens with JRE packages. Besides the fact that not all vendors provide a "JRE" package, the main reason I suggest we keep it out of this is that this action's goal, again, is to set up the JDK environment for CICD purposes. If someone wants to ship a JRE inside a container, they'd probably have to download the package again during the workflow. Or perhaps simply use an existing Docker image as the base for their work. This action works by downloading/caching/extracting the package, and it would be quite troublesome to document (and maintain) where the `tar.gz` was downloaded into for users to reuse that. 

I really see no benefit here to have added complexity, expanded and non-complete matrix of vendors x packages, and using an Action just to download a JRE to be shipped into a Docker image. For the latter, there are many better ways to do that.

bb.

giltene:
@brunoborges 
> 
> Let me address this one first as it directly touches my comment.
> 
> This action's goal is to set up Java - the JDK - for the purpose of a CICD build environment in GitHub Actions for general purpose Java applications. For this reason I recommend we keep it simple.

There are lots of CICD workflows today using jdk+fx and jre packages (several hundreds of each can be found by searching github right now). This sometimes happens when the package type is needed for some functionality, but is also (more often?) driven by wanting to make sure tests actually work with the range of versions and package types a project intends to be "actually work on".

> 
> Given that not all distributions provide a bundle with JavaFX, it would be inconvenient to have this action listing X vendors, but only X-N offering a particular package. For that reason, I support keeping `jdk+fx` out. Users and vendors are still welcome to create their own GH Actions and publish on the Marketplace. The OpenJFX team may as well create their own Action to centralize all JDK+JavaFX bundles in a single action, and then users could use that instead of Setup Java.

I think that as we add support for multiple distros, we must accept that the common denominator available across all distros should not limit the options available for what can be spelled about a requested java setup. This is especially true when considering completely removing (across all distros) the ability to "spell" the need for functionality that is already being used by several hundred existing repos.

Note that the same argument you make about jdk+fx and jre not being available from all distros could then be identically made about java versions and their updates. E.g. not all distros include 6u, or 7u, or 9u, 10u, 12u and 14u, or 14-ea, 15-ea and 16-ea. Some that do no longer update them to the latest update. Some that did no longer keep things that were previously published available "forever". I don't think any of these distro-specific "missing parts" should be used to disqualify a distro from being optionally specified, but it should certainly not be used to stop the associated versions or update levels from being served up by other distros that do support them.

Distro-specific limitations will be a thing. I think that we should accept this and think through what the propper and perhaps common-across-distros "error messages" should exist to make it clear that the thing you are requesting is not available in the distro that you chose, as opposed to the things that you are requesting being a syntax error.

gdams:
> What is the motivation and justification behind the switch of the default from Zulu to AdoptOpenJDK? An argument for having no default would be one thing (i.e. force a choice of a distro and refuse to work without one), but switching the default from one free distro to another seems like a choice that would require at least some significant amount of open debate before taking action.

In my opinion, there are two key reasons here:

1. AdoptOpenJDK/Adoptium is the only entirely vendor natural provider of OpenJDK binaries.
2. I'm not one to focus too heavily on Java surveys but of the 4 or 5 I've seen in the last year, all of them suggest that AdoptOpenJDK is the second most used OpenJDK (Java) binary available today. On those ground, Adopt is at least a likely default choice for many developers.

brunoborges:
As I suggested before (missing link to comment) on the other PR, I think there shouldn't be a default vendor on v2.

While the VM may provide a default JDK version from a specific vendor - today that's Java 8 from Adopt - once a user explicitly adds Setup Action to their workflow, they should provide at the very least two arguments: version and vendor, or a URL for an explicit package to be downloaded and used.

It would be confusing if we end up having a default in the VM that does not match the default in this Action, and such link would make it really hard for us to evolve and upgrade things.

AustinShalit:
Having a default vendor means that there is one less thing for developers to consider when composing a workflow. If there is no default vendor, the action would no longer "just-work" without any inputs.

gdams:
> Having a default vendor means that there is one less thing for developers to consider when composing a workflow. If there is no default vendor, the action would no longer "just-work" without any inputs.

To be clear, existing users are going to have to explicitly bump their setup-java action to v2 so I'm not sure that people are going to be broken. I do agree though.

On the other side, the fewer configuration options to made the action work, the better IMO. Setting a default is unlikely to cause any harm.

brunoborges:
@AustinShalit having this action "working" without any user input is no different than not using this action at all, since the VMs come with JDK installed OOTB.

@gdams we can and should document recommendations in case users are uncertain. But we should also help them understand the JDK landscape. I don't see harm being done if we do require two parameters.

giltene:
There are over 40,000 workflows using setup-java right now. In its current v1.x form it has supported only one (default) distro, and that OpenJDK distro has consistently been free, 100% OSS, and widely used for over seven years, with a wide spectrum of supported java versions and a track re did for uninterrupted updates of all active OpenJDK project versions. With the welcome addition of a number of new, also free, also 100% OSS, also widely used OpenJDK distros over the past couple of years (with e.g. Corretto, Liberica, Adopt, DrangonWell, SapMachine, all coming on the scene), the addition of options for selecting other distros, regardless of whether or not they cover the spectrum of current use, makes perfect sense. Dramatically shrinking the spectrum of use does not, and taking away functionality from existing users does not either.

The silly claims of “neutrality” by one vendor attempting to displace others (one that is clearly the least neutral of the bunch, and the only one slinging this newly invented neutrality mud around) are, frankly, offensive to the multitude of actual OpenJDK contributors that work to actively maintain or develop OpenJDK versions, update them regularly, and keep providing good free binary distros with continued improvements to security and stability. Calling a distro Neutralium does not make it more free, neutral, or deserving, or more free that other distros.

I’d like to steer the discussion around default choices away from arguments that amount to “my distro name has the word OpenJDK in it” and towards actual technical subjects. There are very real functionality and use mode spectrum discussions to be had.

Let me highlight a key part of the spectrum of current use, and let’s remember that some of the most common uses of setup-java is in CICD setups and in regression testing. Let’s also keep in mind that the action is widely used across a spectrum that spans not only team-focused development projects and applications that can control and focus their use cases or versions, but also covers a wide range of publicly available libraries and projects that are then leveraged by many others... This latter group is the one that tends to need a wide spectrum of versions and configurations to test against, and that spectrum includes many of the following:

- OpenJDK 7u, OpenJDK 8u, OpenJDK 11u, and OpenJDK 13u are all actively maintained OpenJDK projects with real users and tons of workflows that build or test against them, both to adapt to the latest updates in each, and (more widely) to identify regressions. The latter generally involves testing against a specific update level and version, while the former involves testing against the latest update available.
- OpenJDK 15-ea OpenJDK 16-ea are currently in active development and there are existing setup-java workflows that seek to ensure their code continue to work when they get released, to identify problems and price feedback early, before those versions GA.
- OpenJDK 6u, 9u, 10u, 12u, and how 14u are all no longer active and do not anticipate additional updates, but all *were* active in the past, and have existing workflows that do regression testing and still expect to work on at least some of those versions. Libraries (and much if the stuff that ends up in maven central) are great example of things that will tend to continue testing on older releases well after the main popular use has moved on to newer versions.

The currently adopted behaviors and available package and version options did not evolve in a vacuum. Let’s not ignore the reasons they exist. The current default supports the spectrum above. The default distro that setup-java currently uses has a long and consistent track record for building a huge spectrum of versions, updates, and package types, rather than arguing that only some are needed, and that approach has led to wide adoption (over 40K workflows in under a year, and growing fast) . I’d be happy to see additional distros step up and start doing the same (rather than argue that the things they don’t choose to build should be taken away from others). But until they do, switching the default to ones that don’t cover the current use spectrum would require one to make some coherent argument for e.g. removing 7u testing capability from the default available to maven central projects, or for not updating to the latest 13u, 15u, or 16-ea update as they come out in OpenJDK. Or for no longer working for a workflows that explicitly test against 8u144, or 6u79, or 12.0.2, or 13.0.3 for regression testing and change isolation reasons (or for e.g. some ALPN version-specific compatibility reasons). Each of these is a current, non-hypothetical use in existing GitHub workflows...

brunoborges:
@giltene I am advocating for v2 to _not_ have a default vendor. Are you in favor of that?

Users would have to pick a vendor and a version:

```yaml
- name: Set up JDK 11
      uses: actions/setup-java@v2
      with:
        java-version: 11
        java-distro: zulu
```

Frankly, I am not following why so much debate for an action that should be able to cover 80% of the cases in a very simplistic way, and for the other 20%, developers can easily set up JAVA_HOME using wget/curl.

giltene:
> @giltene I am advocating for v2 to _not_ have a default vendor. Are you in favor of that?

Switching the default (which is the initial proposal, and the main thing I am responding to) and eliminating the default are two separate things. You are correct that second if chosen, the latter will obviously make the former moot. But unless serious momentum behind removing the default altogether starts forming, I'm focusing my discussion on the first one. 

As for removing the default: I think that it is a matter that warrants separate discussion. I can see some merit in it, and a logical argument that perhaps it should have been that way from the start, and that e.g. even when only one distro was supported with ways of spelling out JDK version and package requirements (and matrixes around them), the distro option should have been required. I'm on the fence on that one [i.e. what would the "right way to start" have been]: the simplicity of not spelling out the distro probably helped, and is clearly natural. But it also led us to the "oh, there is more than one OpenJDK distro, which one do you mean" discussions and to the sort of "this is the wrong default" arguments that inevitably arise from choice. 

But hindsight may not serve us well now. There are 40K+ workflows happily using the current scheme, and there are plenty of options (in the form of other actions) available for the small minority who are unhappy with its defaults right now. Many of the 40K+ workflows will likely break (or worse, choose to stick much longer to 1.x) if forced to add or change options to make their current workflows continue to work.

The cleanest way to move people to new versions and capabilities is to avoid breaking changes wherever possible. That's not to say that breaking changes should never be made, but when you do make them, there should be some real reason and benefit to existing users (or to new users that were unable to use stuff before the change is introduced) for taking that action. When choosing to introduce a breaking change, trying to minimize the % of existing cases that would actually break should be a key design concern as well.

The options we are discussing are:

a) Eliminating the default (and requiring an extra option) will create a breaking change for 100% of workflows that would switch to 2.x.

b) Changing the default (as suggested at the top of this discussion) will create a breaking change for some unpredictable portion of workflows that currently make use of anything that AdoptOpenJDK does not currently offer (this currently includes 7u, 8u versions before some update level, 13 updates after some update levels, 15-ea or 16-ea, 6u, 9u, 10, jdk+fx packages, etc., each of which has actual uses in current workflows)

c) Keeping the current default and adding the distro option, along with expanding the distro option to cover not juts Zulu and AOJ, but also Corretto, Liberica, Dragonwell, Oracle OpenJDK, SAP Machine, etc., will address the actual needs stated by people for having alternate distros, without forcing a breaking change on any of the current 40K+ workflows that use 1.x, and without forcing anyone to break up their matrixes or finding new hardcoded URI replacements to their current working matrix setups...

to me, (c) seems the obvious right path. I'm waiting for an "(a) or (b) would be better because of X, and X is worth breaking 40K existing workflows to achieve." argument.

> 
> Users would have to pick a vendor and a version:
> 
> ```yaml
> - name: Set up JDK 11
>       uses: actions/setup-java@v2
>       with:
>         java-version: 11
>         java-distro: zulu
> ```
> 
> Frankly, I am not following why so much debate for an action that should be able to cover 80% of the cases in a very simplistic way, and for the other 20%, developers can easily set up JAVA_HOME using wget/curl.

The removal of the need to point test workflows to specific URIs is one of the key drivers of setup-java's success IMO, and is the historical use mode in CICD systems. For the same reason that we usually look to avoid spelling out code dependencies in Maven or Gradle with URIs for specific jars. 

The ability to cleanly spell out and run a test matrix across e.g. [7, 7u242, 8, 8u232, 11, 11.0.4, 13, 13.0.4] (and add 15-ea, and 16-ea to that for future-proofers) is here today, and is much wider than 20% IMO. Most projects that produce code used by others (libraries, stuff on maven central, etc.) need this sort of test matrix. It is the likely best practice for projects that are looking to both test for regressions introduced in their own code (and thus should be testing on fixed JDK update levels) and for regressions or issues introduced in new updates to JDKs (and thus should be testing on the latest update of each).

giltene:
Here is a suggested [potential] staged approach:

1. Actually add a distro option
1.1 enable multiple distros (aim for "multiple" being >2).
1.2 work through the issues that arise as people use the new distro option (e.g. what to do when a version or package type being requested is not available in the distro being requested. Providing clear error messages that do not suggest the syntax or value for the requested version or package is wrong, but instead making it clear that the version/package combination requested does not exist in the distro requested).
1.3 seek feedback, watch adoption, fix problems.

2. Once we have a working and somewhat adopted multi-distro stuff working, open a discussion about the potential next steps:
2.1 Should the default be dropped, and the (now provably working for people) distro option made mandatory?
2.2 If the default should not be dropped, should it be changed to a distro that is different than the one currently used as the default?

Making an actual choice on 2.1 and 2.2 without actually having a working distro option that people can use, give feedback on, and which we can do work to improve through that feedback seems like a recipe for more hind sight discussions later on...

filiphr:
I'd like to chime in as a maintainer of 2 different Java OSS projects that are using this action to verify that our projects are working on multiple Java versions.

First I have to admit that before the discussion here and in #13, I was not aware of the different way the different JDK distributions work. Thanks to everyone for contributing to these discussions, I learned quite a lot.

One of the key things for me as a maintainer of an OSS project is the easy way of setting up my CI/CD and having the trust that it will continue working without a breaking change, especially when I choose a different version such as 15-ea, 16-ea as a JDK version. If a vendor chooses to remove some JDK from being available for download this would mean that my CI/CD will break and I will have to rush and find the time to fix it, even though nothing in my project changed. From my current understanding it seems that only Zulu offers a perpetual URL for downloading those JDKs. Therefore I think that Zulu is a good option for multitude of OSS projects.

The fact that you want to switch to v2, only because you want to change the default option of the JDK means that the users of this action will be split between v1 and v2 for quite some time. A lot of OSS maintainers are limited in their time they have to spend on the projects, they are not backed by companies and are doing this in their own free time. Therefore, I guess a lot of them will stay on v1 for quite some time, or something will break for them if they are using `@main` or `@master`.

Having said all of this in my opinion as an OSS maintainer I think that the approach proposed by @giltene ( 1. Add distro option, 2. Decide if changing the default is OK or decide if distro should be mandiatory) is a really good one. In order to get the best feedback and adoption I would even say that "adding distro option" can be done in v1 safely.

I would also like to ask to add that this ADR should mention the different rules a specific distro provides their versions. e.g. currently AdoptOpenJDK does not have the "X-ea" and pre 8 (this is actually a specific 8u252) builds, Oracle OpenJDK removes the X-ea builds once the X version is released. The reason for me mentioning this is that in my opinion as an OSS maintainer, I would like to be able to test with the versions I have defined in my action and expect it to work without the need to do anything specific. If I choose an explicit vendor then it is completely expected for something to stop working if that distro no longer provides the version I need.

davidkarlsen:
can anything be done to move this forward - as of today I have to depend on a fork for adoptopenjdk support - and would prefer using the official action.

eddumelendez:
LGTM! I can not wait for it or contribute in some way.

davidkarlsen:
@chrispat @madhurig PTAL?

eddumelendez:
FYI I have created a github action for `sdkman` which allows to install any java distribution and taking advantage of `setup-java` to pass the file to set the proper jdk.

chrispat (approved):

giltene:
To help move this forward, I'd like to re-iterate my suggestion in https://github.com/actions/setup-java/pull/97#issuecomment-680223560 : Lets focus V2 on adding actual ability to support multiple distros in a non-breaking way for existing workflows, and remove the changes to the defaults from the goals (and punt discussion about them to a later time).

konradpabjan:
This ADR stalled a bit because we haven't had time to pick this up. We should be revisiting this soon. @alenasviridenko will be helping out to move this forward

I've made some small clarifications to the ADR based off of some of the feedback here and some internal discussions that we've had. For `v2` the new `distribution` input parameter will be required and we will **not** be choosing a default for users.

This will causes workflows to fail for users that are pinned to `@main` once our changes are merged in. Users will have to go and update their YAML, it should be easy one line fix but breaking users is never fun that we try to avoid (even though we tell users pinning to `@main` has it's risks).

There are also a few comments about first adding a distro option and then changing the default behavior. We are going to be avoiding this. The main reason is because breaking changes should only be introduced when going to a new major version (`v1` to `v2`) as we don't guarantee backwards compat. If there is a time to break compatibility it's now. We also don't want to do something like introduce just a distro option with `v2` (or the current version) and then change the default behavior in an upcoming version like `v3`.

giltene:
FYI, work towards a generic means of discovering bits for a distribution is being done within [foojay.io](https://foojay.io), and will likely soon be exposed as an API useable from within the setup-java implementation. The API will cover the same sort of options setup-java now supports for specifying things (version specification in server format (extended to support [JEP322](https://openjdk.java.net/jeps/322)), package type), as well as specifying a distribution and a target platform.

Expected distros supported out of the gate will likely include Zulu, AOJ, Corretto, SapMachine, Liberia, Dragonwell, and Oracle OpenJDK (Adoptium will likely be added as soon as it starts making bits available). The community need for such an API has become evident in multiple areas, with actions/setup-java being one example out of many: IDEs and various other build toolchains share this need as well. Beyond being able to locate bits by specification as is the need in the setup-java use case (and in typical CI/CD setups), the API will provide natural flow for discovering e.g. available java versions and update levels, and the distributions that support a given version, update level, and platform (as is the likely need in e.g. an interactive IDE workflow, where a developer may browsing for bits for a given java version or update level, but may not know which distributions have such bits available). Given it's core purpose, [foojay.io](https://foojay.io) seems like a good place to host and maintain such an API. 

Once the foojay.io distro discovery API is up, we'll be able to address https://github.com/actions/setup-java/issues/13, https://github.com/actions/setup-java/issues/68, https://github.com/actions/setup-java/issues/69, https://github.com/actions/setup-java/issues/70, https://github.com/actions/setup-java/issues/71, https://github.com/actions/setup-java/issues/72 with a single, common logic path that will work for all, and will even allow for the inclusion of additional (future) distros with no further code changes. We'll be able to add a `distribution` option as a non-breaking change in setup-java at that point, and it can simply be added to a 1.x version as an incremental improvement. Regardless of what 2.x chooses to do about the default behavior, it would make sense to use the same logic  in 2.x.

docs/adrs/0000-v2-setup-java.md
```diff
@@ -10,13 +10,13 @@ Status: Proposed
 
 - Zulu and AdoptOpenJDK aren't the only distributions of Java though. Other providers include Oracle OpenJDK or Amazon Corretto and ideally it would be nice to support downloading Java from all providers.
 
-- GitHub Actions virtual environments install and default to AdoptOpenJDK builds of OpenJDK. `setup-java` should align to the same behavior.
```
madhurig: typo: `download`

docs/adrs/0000-v2-setup-java.md
```diff
@@ -70,5 +72,5 @@ The `setup-java` action has some logic that creates a `settings.xml` file so tha
 
 - Users will have more flexibility and the freedom to choose a specific distribution that they would like (AdoptOpenJDK builds of OpenJDK in addition or Zulu builds of OpenJDK)
 - `setup-java` will be structured in such a way that will allow for more distributions to be easily added in the future
-- A large subset of users pin to `@main` or `@master` instead of to a specific version (`v1`). By introducing a breaking change by switching from `Zulu` to `AdoptOpenJDK` as the default, certain existing workflows for users might break
+- A large subset of users pin to `@main` or `@master` instead of to a specific version (`v1`). By introducing a breaking change that now requires a distribution to be specified, many users will have their workflows fail until they go and update their yaml
```
madhurig, line 75: We should update that only a small % of users pin to main and that we don't recommend it.

madhurig (approved):

docs/adrs/0000-v2-setup-java.md
```diff
@@ -0,0 +1,76 @@
...
+  with:
+    java-version: '11'
+    distribution: adoptopenjdk
+```
+
+## Default Behavior
+
+There will be no default distribution that we pick for the user. Users will have to specify a distribution in their YAML or else the action will fail.
+
+Requiring a default version will break users that are pinned to `@master` or `@main` as they will have no `distribution` specified in their YAML. Users pinned to `v1` will be uneffected. This change is meant to not be backward compatible and it is acceptable to change the default behavior because a new major version will be released alongside these changes.
```
maxim-lobanov, line 31: We don't really care about `@master` since we won't update it. Only `main`

docs/adrs/0000-v2-setup-java.md
```diff
@@ -0,0 +1,76 @@
...
+# Decision
+
+## New input
+
+A new required input parameter (titled `distribution`) will be added to `setup-java` that will allow users to specify the distribution that they would like to download
+
+```yaml
+  with:
+    java-version: '11'
+    distribution: adoptopenjdk
```
maxim-lobanov, line 24: rename to `adoptium`

docs/adrs/0000-v2-setup-java.md
```diff
@@ -0,0 +1,76 @@
...
+
+`setup-java` should be structured in such a way that will allow the open source community to easily add support for extra distributions.
+
+Existing code will be restructured so that distribution specific code will be easily separated. Currently the core download logic is in a single file, `installer.ts`. This file will be split up and abstracted out so that there will be no vendor specified logic. Each distribution will have it's own file under `src/distributions` that will contain the core setup logic for a specific distribution. 
+
+```yaml
+ ∟ src/
+    installer.ts # core installer logic
+    distributions/
+        ∟ adoptOpenJDK.ts # adoptOpenJDK specific logic
```
maxim-lobanov, line 43: `adoptium`

konradpabjan:
Merging in the ADR! Some CI is failing but it's unrelated to the changes in this PR. The tests are being overhauled significantly with `v2` anyway

maxim-lobanov (approved):

giltene:
Where can I find tueclwtest version of the actual ADR? The link to a rendered version at the top of this PR seems to be broken.

PR #104, opened 2020-09-29
Swap to Environment Files

thboop:
The newest toolkit version uses [Environment files](https://github.com/actions/toolkit/blob/main/docs/commands.md#environment-files) to communicate with the Runner. This pr moves towards the new way of communicating with the runner and updates the toolkit version if appropriate.


konradpabjan (approved):

srl295:
For the record, this landed in `actions/setup-java@v1.4.3`

(I was mystified as to why I was getting `set-env` warnings when I didn't use `set-env`…directly.)

PR #106, opened 2020-10-03
Bring Ubuntu current in README

gliptak:


README.md
```diff
@@ -69,7 +69,7 @@ steps:
 ```yaml
 jobs:
   build:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-20.04
```
AustinShalit, line 72: What do you think about switching it to latest?

AustinShalit (changes requested):

README.md
gliptak: both intents are valid. for stricter reproducible build, one might choose to set the version.

let me know if I'm to update the PR

AustinShalit: For a readme, I think `latest` is fine.

gliptak: updated

AustinShalit (approved):

gliptak:
@AustinShalit anything else before this can be merged?

AustinShalit:
I don't think so but I am just another person like you. I have no power here.

PR #132, opened 2021-03-08
actions/setup-java@v2 - Support different distributions

maxim-lobanov:
**Description:**
This pull-request reworks installer approach and create abstract base that can be used for adding new Java distributions. Also this pull-request adds support for Azul and Adoptium (AdoptOpenJDK) distributions.
These changes are going to be released as `actions/setup-java@v2` so breaking changes are expected.

Note: we have decided to split changes to a few pull-requests to simplify review process:
- Installers: (this pull-request) (workflows are temporary removed, they are added back in next PR)
- Tests: https://github.com/actions/setup-java/pull/133
- Docs: https://github.com/actions/setup-java/pull/134
- Related ADR: https://github.com/actions/setup-java/pull/97

A few technical notes:
1) Added `distribution` input without default value.
2) Added support for two distributions: `zulu` (Azul OpenJDK), `adoptium` (Adopt OpenJDK)
3) The approach for searching Azul Java versions was reworked from scratch. Previously, actions/setup-java@v1 parsed html page `https://static.azul.com/zulu/bin/` and use regex to find Java versions.
We have reworked it to use Azul API: https://app.swaggerhub.com/apis-docs/azul/zulu-download-community/1.0. **Potentially, it could cause some differences in version resolving** but we have tested all major use-cases and didn't find any issues.
4) Use Adopt API V3 https://api.adoptopenjdk.net/swagger-ui/ to get the list of versions. Unfortunately, API has very hard limitation (20 items per page) so need to paginate pages.
5) Use the following tool naming in toolcache: `Java_${distribution}_${packageType}` to make sure that we can cache different vendors side by side.
Also, adds `-ea` postfix for version in toolcache to distinguish stable and unstable versions (in V1 - it could cause issues)
5) Bumped all dependencies to latest versions (no risk since it is major release)
6) We have decided to make `distribution` and `java-package` properties optional to simplify use-cases when task is used for setting up Maven publishing (settings.xml), customers don't really need to spend time on Java setup. So if `distribution` and `java-version` properties are specified - Java will be installed. Otherwise - only settings.xml will be created. Also both `distribution` and `java-package` should be specified together (if only one is specified - error will be raised)
7) Java versions are stored in tool-cache with 3 digits like (11.0.3) despite the fact that Java versioning is 4 digits (the fourth is `build`). The same behavior was in V1. It could cause some confusing but neither V1 nor V2 support specifying versions with 4 digits in input - so it shouldn't be an issue. Also, looks like `toolkit/toolcache` doesn't support 4 digits versions because of `semver.clean` invocation.
8) Minor refactoring of auth / gpg code: moving methods to the separate files, replace `console.log` with `core.info`, apply styling. No major changes there.
9) Remove `version` input in favor of `java-version` (V1 supported both for backward compatibility)


src/distributions/base-installer.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  protected abstract downloadTool(javaRelease: JavaDownloadRelease): Promise<JavaInstallerResults>;
+  protected abstract findPackageForDownload(range: semver.Range): Promise<JavaDownloadRelease>;
+
+  public async setupJava(): Promise<JavaInstallerResults> {
+    let foundJava = this.findInToolcache();
+    if (foundJava) {
+      core.info(`Resolved Java ${foundJava.version} from tool-cache`);
+    } else {
+      core.info(`Java ${this.version.raw} is not found in tool-cache. Trying to download...`);
```
konradpabjan, line 37: nit: `was not found in the tool-cache`

src/distributions/base-installer.ts
```diff
@@ -0,0 +1,115 @@
...
+    let stable = true;
+
+    if (version.endsWith('-ea')) {
+      version = version.replace(/-ea$/, '');
+      stable = false;
+    }
+
+    if (!semver.validRange(version)) {
+      throw new Error(
+        `The string '${version}' is not valid SemVer notation for Java version. Please check README file for code snippets and more detailed information`
```
konradpabjan, line 106: Nit: `The string '${version}' is not valid SemVer notation for a Java version. Please check the README file for..`

src/distributions/local/installer.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  public async setupJava(): Promise<JavaInstallerResults> {
+    let foundJava = this.findInToolcache();
+
+    if (foundJava) {
+      core.info(`Resolved Java ${foundJava.version} from tool-cache`);
+    } else {
+      core.info(
+        `Java ${this.version.raw} is not found in tool-cache. Trying to unpack JDK file...`
```
konradpabjan, line 25: `Java ${this.version.raw} was not found in the tool-cache. Trying to ...`

Same thing line 34. `JDK file was not found...`

src/distributions/local/installer.ts
```diff
@@ -0,0 +1,78 @@
...
+        javaPath = path.join(javaPath, MACOS_JAVA_CONTENT_POSTFIX);
+      }
+
+      foundJava = {
+        version: javaVersion,
+        path: javaPath
+      };
+    }
+
+    core.info(`Setting Java ${foundJava.version} as default`);
```
konradpabjan, line 65: nit: `as the default`. Sounds a little bit better

konradpabjan (commented):
Looking great! Did an initial sweep, just a few things

src/distributions/adoptium/installer.ts
```diff
@@ -0,0 +1,146 @@
...
+      const requestArguments = `${baseRequestArguments}&page_size=20&page=${page_index}`;
+      const availableVersionsUrl = `https://api.adoptopenjdk.net/v3/assets/version/${encodedVersionRange}?${requestArguments}`;
+      if (core.isDebug() && page_index === 0) {
+        // url is identical except page_index so print it once for debug
+        core.debug(`Gathering available versions from '${availableVersionsUrl}'`);
+      }
+
+      const paginationPage = (
+        await this.http.getJson<IAdoptiumAvailableVersions[]>(availableVersionsUrl)
+      ).result;
```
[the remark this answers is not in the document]
maxim-lobanov: I think http-client will report error message and status code in case of failure (if statusCode is not successful).
Also I have checked Adopt swagger API, http-client should handle all status codes properly (the same for Azul)

Currently, we don't have retry logic and try-catch on task level and I am not sure that it makes sense to introduce it.

giltene:
FYI, the Disco API (https://github.com/foojay2020/discoapi) is coming along nicely, with multiple tools starting to use it. It provides a sweeping solution for the "find a JDK" task across multiple distros, and an actions/setup-java PR is coming soon for using it directly, replacing the current distro-specific code, but keeping all existing functionality. This will likely overlap with this PR.

See some FOSDEM discussion from last month: https://fosdem.org/2021/schedule/event/discoapi/

AlenaSviridenko (approved):

gdams (approved):

joschi:
> FYI, the Disco API (https://github.com/foojay2020/discoapi) is coming along nicely, with multiple tools starting to use it.

Where might I have seen that before?
https://github.com/actions/setup-java/issues/13#issuecomment-634930137

konradpabjan (approved):

konradpabjan:
Down the road we can consider utilizing the Disco API or any other tools to expand support for more distributions. With the initial v2 preview and release we're primarily focusing on just expanding support for Adoptium and ironing out any bugs that might come up since that was the biggest ASK from users.

MaksimZhukov (approved):

giltene:
See https://github.com/actions/setup-java/pull/134#issuecomment-795852412 about the meaning of using "adoptium" as a distro name.

giltene:
How much testing of the new Zulu-getting logic has been done to see if it covers existing working setups? There is good corpus of version tests at https://github.com/giltene/jdkTestingMatrix that can be used to see if everything still works right with the new logic...

giltene:
I'm confused by the switch to making "distribution" property optional. That seems to [strongly?] contradict the "no default" arguments of before. If "distribution" is optional, there should be a default distro, and we should open that discussion up again [which I think would be a good idea]. If there its no default distro, "distribution" should not (and cannot) be optional.

gdams:
> I'm confused by the switch to making "distribution" property optional. That seems to [strongly?] contradict the "no default" arguments of before.

In the near future, we may have the option to add a default. Right now I see no issue with forcing users to choose which JDK to use going forwards. If anything it encourages people to make a conscious decision.

giltene:
> Down the road we can consider utilizing the Disco API or any other tools to expand support for more distributions. With the initial v2 preview and release we're primarily focusing on just expanding support for Adoptium and ironing out any bugs that might come up since that was the biggest ASK from users.

I haven't seen much of a user ask for Adoptium. It would be surprising to see one since there are no Adoptium builds available yet.

On the other hand the ask for a choice between multiple existing distributions is very real. And there are several available and being asked for.

giltene:
> > I'm confused by the switch to making "distribution" property optional. That seems to [strongly?] contradict the "no default" arguments of before.
> 
> In the near future, we may have the option to add a default. Right now I see no issue with forcing users to choose which JDK to use going forwards. If anything it encourages people to make a conscious decision.

Ok, but that would mean the "distribution" property is NOT optional. At the top of this page (tech note #7), it says the opposite. Let's change that then.

maxim-lobanov:
@giltene , `distribution` property is mandatory when `java-version` is specified.
When customer doesn't need to install Java at all and use task to only generate settings.xml and configure signing, he will be able to skip both `java-version` and `distribution` properties. But when `java-version` is specified, `distribution` is mandatory. Otherwise, task will fail.
See related code: https://github.com/actions/setup-java/blob/implement-v2-installers/src/setup-java.ts#L17

> We have decided to make distribution and java-package properties optional to simplify use-cases when task is used for setting up Maven publishing (settings.xml), customers don't really need to spend time on Java setup. So if distribution and java-version properties are specified - Java will be installed. Otherwise - only settings.xml will be created. Also both distribution and java-package should be specified together (if only one is specified - error will be raised)

maxim-lobanov:
@giltene , thank you for sharing https://github.com/giltene/jdkTestingMatrix.
We have validated the most of versions via https://github.com/actions/setup-java/blob/implement-v2-tests/.github/workflows/e2e-versions.yml
I will try to validate V2 via your matrix tomorrow

giltene:
> @giltene , `distribution` property is mandatory when `java-version` is specified.
> When customer doesn't need to install Java at all and use task to only generate settings.xml and configure signing, he will be able to skip both `java-version` and `distribution` properties. But when `java-version` is specified, `distribution` is mandatory. Otherwise, task will fail.

This is an inconsistent approach. We should either require `distribution` to be specified explicitly in all cases, or make it optional in all cases (which means there is an implied default distribution). The argument for making it optional for some uses, like ones where no java version is specified [a use case that does not currently exist for this action], is not different than making it optional when e.g. a java version is specified but the update level or build detail was not. We either accept that all uses of this action will start requiring the specification of a distro, or we don't.

On a related note, changing java-version to be optional seems like a very (very!) bad idea. Just think of the configuration control nightmare that would be introduced (and encouraged) by this. Leaving the major java version selection up to "whatever happens to be around today" is a sure way to make common things that work today break tomorrow. It's a bad habit that none of the current (teams of thousands of happy and successful) users of the seta-java action have, but one that that would pop up immediately when enabled. 

The main role of the setup-java action is to set up java. Not to set up maven files or gradle files or gpg stuff. Having those other things ride along with setup-java for convenience makes sense to some degree (separate discussion), but changing the main java version setup action to not do its primary job in order to accommodate a convenience path of a side effect is wrong. Maybe there should be a separate maven action to deal with such e.g. maven convenience asks?

filiphr:
>The main role of the setup-java action is to set up java. Not to set up maven files or gradle files or gpg stuff. Having those other things ride along with setup-java for convenience makes sense to some degree (separate discussion), but changing the main java version setup action to not do its primary job in order to accommodate a convenience path of a side effect is wrong. Maybe there should be a separate maven action to deal with such e.g. maven convenience asks?

As an open source maintainer I have to agree with @giltene here. I was not even aware that the "setup-java" action does things like maven / gradle settings configuration. This is totally unexpected. Making the `java-version` optional is a recipe for disaster for already strained open source maintainers. The default containers where GitHub actions are running have some kind of a java version available. This means that compiling might work, but in an unexpected way. 

I would ask you to reconsider this approach.

maxim-lobanov:
@giltene, thank you for your input on distribution naming. We have renamed `adoptium` to `adopt` across all code. It looks shorter and more familiar for customers.

Also, we have validated V2 against your matrix: https://github.com/giltene/jdkTestingMatrix, thank you for sharing it, really great coverage.
Everything works as expected except a few versions that are missed in Azul API for some reason but available on this HTML page https://static.azul.com/zulu/bin/ :
- JDK 15.0.0-ea.2 (Azul API provides only 15.0.0-ea.11 and higher)
- JDK 16.0.0+5 (Azul API provides only 16.0.0+8 and higher)
- JRE macOS 11.0.6, 11.0.7, 11.0.8, 11.0.9, 8.0.242, 8.0.252, 8.0.262, 8.0.265, 8.0.272, 8.0.275

maxim-lobanov:
@giltene , @filiphr , agree that having extra functionality related to Maven / gpg in setup-java task looks confusing but we have to live with it since it is already implemented and we don't have plans to drop it in V2.
We don't see any problems with making these inputs optional because it is pretty clear that `java-version` should be specified in order to set up Java. Also, README and other documentation says about it. If user doesn't specify Java version in setup Java task - what version they are expecting to get?

On the other hand, setting singing / GPG only (without changing Java version) is absolutely valid use-case. And it looks strange to force user to specify `java-version` and `distribution` inputs in YAML file when he doesn't really need to change Java version. Also Java installation takes additional time - it doesn't make sense to spend user's build time on installing it.

konradpabjan:
The main reason why `setup-java` has Maven functionality in it is so that it's easier to publish to GitHub package registry. When the initial maven publishing feature was added there was a pretty lengthy debate (and I was very much against adding the functionality to `setup-java`) but we ultimately decided to do it so that publishing would "just work" without any extra configurations or any other actions. I agree that it somewhat distorts the single purpose intent that `setup-java` initially was meant to provide (and that most users think it provides) but we ultimately decided that the pros outweight the cons.

One solution that was floating around was to create a separate `setup-maven` or `setup-gradle` package but we don't have the bandwidth to support yet another action and the maintenance burden would also be very high so we just extended `setup-java`

Before we started work on the `v2-preview`, we had some discussions regarding what to do with the existing Maven publishing functionality and the general consensus was to just leave it as is. Don't prioritize any new features for it, don't break what we currently have, but keep the current functionality as is because we still want publishing to GitHub packages to "just work".

Given that we're maintaining the existing Maven publishing functionality, allowing people to use it without specifying any input is a valid scenario considering our hosted runners already have certain Java versions pre-installed and ready to go.

action.yml
```diff
@@ -4,22 +4,21 @@ description: 'Set up a specific version of the Java JDK and add the
 author: 'GitHub'
 inputs:
   java-version:
-    description: 'The Java version to make available on the path. Takes a whole
-       or semver Java version, or 1.x syntax (e.g. 1.8 => Java 8.x).
-       Early access versions can be specified in the form of e.g. 14-ea,
-       14.0.0-ea, or 14.0.0-ea.28'
-    required: true
+    description: 'The Java version to set up. Takes a whole or semver Java version. See examples of supported syntax in README file'
+    required: false
```
sormuras, line 8: Nah! Please, require users of the `setup-java` action to commit to specific version of Java.

At least a feature release number, like: `8`, `11`,  ... `17`

maxim-lobanov: - https://github.com/actions/setup-java/pull/132#issuecomment-795918486
- https://github.com/actions/setup-java/pull/132#issuecomment-796672020

sormuras: I read and understood the comments. Yet, I don't agree with your conclusions.

I'll reply in the main thread.

sormuras:
Perhaps too late for `setup-java@v2`, may be a good starting point for `setup-java@v3`, but here's what I (_as an early adopter of GH Actions, writer of the [hello-world-java-action](https://github.com/sormuras/hello-world-java-action) and maintainer of [`install-jdk.sh`](https://github.com/sormuras/bach/tree/master#install-jdksh) used by Travis CI_) think about `setup-java`.

Make users think about and commit to a Java version

Compare how GH users select an OS: `xyz-latest`, or `xyz-VERSION` - there's no plain `xyz`. Users must explicitly state that they either don't care for the actual version (gimme "latest-and-greatest", I'll care about problems when they occur) or that they want exactly `xyz-VERSION` OS (perhaps, they make use of pre-installed software of the OS image).

Make one thing good

Also, splitting `setup-java`  into dedicated `setup-java`, `setup-maven`, `setup-xyz` ... actions is a good idea! Focus on setting up Java, nothing more, nothing less - and let the other projects provide their own customized setup actions, building on a stable and clean `setup-java` action. I bet, Maven committers (as myself) or Gradle devs and others are eager to maintain a dedicated GH for their tool(ing).

Another fine aspect of making a dedicated `setup-java` action is to leverage pre-installed JDK(s) for the sake of speed. For example, if a user requests a `java-version: 11` (and doesn't specifiy any other arguments) and the current image has a matching JDK pre-installed: use it! Zero time lost. If there's no matching version, install (emitting a note that `11` is pre-installed and could be used without overhead) and use that installation. Side-effect: if 90% users request `17` some day, pre-install `17`.

sormuras:
The initial description of this PR doesn't mention the `jdkFile:` input variable supported by `setup-java@v1`.

In https://github.com/sormuras/download-jdk I am making use of it (right next the good `java-version` input).

Will `setup-java@v2` support the `jdkFile:` input unchanged?

maxim-lobanov:
@sormuras , thank you for feedback.

1) `V2` will be able to use pre-cached versions on Hosted images if they exist. We are doing related changes on images side (all platforms). Example: https://github.com/actions/virtual-environments/pull/2883 . So pre-cached versions will be switched without latency.
2) Yes, `jdkFile` is still supported in V2 with minor syntax change. See https://github.com/actions/setup-java/pull/134/files#diff-e8f077830705686ec3300dbcec8ee6ac6345ae9a2965508ff3abc4146d520bf8R77

giltene:
I don't see a problem with maven uses needing to specify the java version that maven would use when using the setup-java action to do maven things, including publishing. I frankly can't see any value in committing it, in even in the most trivial use cases. I see the mandatory specification of a major java version as a necessity (both for configuration control reasons, and since maven behavior itself can be sensitive to java versions), and see the ability to do *anything* with setup-java without specifying the java version as introducing a huge negative. E.g. mvn build can [and will] go very badly wrong, in both silent and non-silent ways, if used with changed-at-some-random-point-in-the-futrure java versions. And this will happen (and propagate by example to thousands of places) the minute "simple" uses of maven start working (temporarily) without java-version specified.

I strongly recommend that java-version remain a required property, like it is currently in v1.x.

giltene:
@maxim : On the caching subject: can we discuss this in some more detail? [and is there a separate thread to do that on?] E.g. will caching check that the cached versions are actually identical (e.g. have the same checksum or hash) as the thing that would have been downloaded had a cached version not been found? There are many update scenarios (e.g. whenever quarterly OpenJDK updates are rolled out, and whenever out-of-cycle critical updates are released, both of which are frequent enough to not ignore) where stale cache decisions could be very problematic, and such decisions an be avoided if the cache logic verifies the checksums of what is available in the cache against the checksums of what would be downloaded.

We are adding hash/checksum information capability in the Disco API for this very reason, with the intent to enable all distros to provide hash value for their download URIs in order to facilitate proper caching functionality.

maxim-lobanov:
> @maxim : On the caching subject: can we discuss this in some more detail? [and is there a separate thread to do that on?] E.g. will caching check that the cached versions are actually identical (e.g. have the same checksum or hash) as the thing that would have been downloaded had a cached version not been found? There are many update scenarios (e.g. whenever quarterly OpenJDK updates are rolled out, and whenever out-of-cycle critical updates are released, both of which are frequent enough to not ignore) where stale cache decisions could be very problematic, and such decisions an be avoided if the cache logic verifies the checksums of what is available in the cache against the checksums of what would be downloaded.

@giltene , caching logic will work the same way like it works for other tools (go, node, python, haskell) - compare version only, not checksum.
Images will contain latest version of every LTS and will be updated with 1-week delay.
First, action tries to find version that satisfies input locally. If it is not found locally, it will download and install version in runtime.
If you specify exact version like `11.0.10+9` but image contains only `11.0.10+8`, then `11.0.10+9` will be downloaded in runtime.

maxim-lobanov:
@sormuras , we have validated `actions/setup-java@v2` with your repository: https://github.com/dmitry-shibanov/download-jdk/actions/runs/645961061 - it works as expected

ibauersachs:
@maxim-lobanov I'm a somewhat heavy user of setup-java and cannot think of a _single_ reason not to specify `java-version`. The Maven/Gradle setup is a convencience (that I like), but if that was the _only_ thing I'd need, I would definitely not use setup-java. Please, make the `java-version` and `distribution` properties mandatory!

src/auth.ts
```diff
@@ -1,26 +1,52 @@
...
-    `creating ${SETTINGS_FILE} with server-id: ${id};`,
-    'environment variables:',
-    `username=\$${username},`,
-    `password=\$${password},`,
-    `and gpg-passphrase=${gpgPassphrase ? '$' + gpgPassphrase : null}`
+  core.info(
+    `Creating ${SETTINGS_FILE} with server-id: ${id};
+     environment variables:
+     username=\$${username},
+     password=\$${password},
```
ibauersachs, line 48: Why are these secrets logged? I know GitHub Actions masks repo/org secrets in the output, but it's a really bad idea to rely on this and to log credentials in general.

src/distributions/adopt/installer.ts
```diff
@@ -0,0 +1,145 @@
...
+import { MACOS_JAVA_CONTENT_POSTFIX } from '../../constants';
+import { extractJdkFile, getDownloadArchiveExtension, isVersionSatisfies } from '../../util';
+
+export class AdoptDistribution extends JavaBase {
+  constructor(installerOptions: JavaInstallerOptions) {
+    super('Adopt', installerOptions);
+  }
+
+  protected async findPackageForDownload(version: string): Promise<JavaDownloadRelease> {
+    const availableVersionsRaw = await this.getAvailableVersions();
```
ibauersachs, line 20: Do understand this correctly? Are you downloading the _entire_ available list of AdoptOpenJDK binaries just to later filter if the version in the action matches one of the binaries?
If so, this is insane. The API provides functionality to e.g. get the latest 11-ga binary. Use this, or this action (or the entire GitHub IP range) will likely get banned for too many requests.

src/auth.ts
maxim-lobanov: I don't have context why this logging was initially added in V1 but actually, it doesn't even log secrets - it logs names of variables where these values are stored. 
See action usage: https://github.com/actions/setup-java/pull/134/files#diff-e8f077830705686ec3300dbcec8ee6ac6345ae9a2965508ff3abc4146d520bf8R154
User doesn't pass secrets to action, it passes names of variables where secrets are located.

Anyway, I agree that it doesn't make much sense to log this info into logs. I have cut the log a bit

src/distributions/adopt/installer.ts
maxim-lobanov: Yes, it is necessary to download entire list to resolve customer's input correctly.
It is not enough to get `latest 11-ga binary` because user can specify version as `11` or `11.0` or `8.0.232` or even `8.0.282+8`. This task supports [SemVer](https://semver.org) notation for input so we need the full list of available versions to resolve input properly.

Actually, getting all versions is not a problem. Adopt just provides about ~50 versions that is pretty small list. It takes less than 0.1 seconds to retrieve it. It is not a problem even for Azul that provides wider list of available versions.

We don't see any problems with it

maxim-lobanov:
Thank you for providing your input, I have pushed PR update to mark `java-version` and `distribution` properties as required.

src/distributions/adopt/installer.ts
ibauersachs: > Yes, it is necessary to download entire list to resolve customer's input correctly.
> It is not enough to get `latest 11-ga binary` because user can specify version as `11` or `11.0` or `8.0.232` or even `8.0.282+8`. This task supports [SemVer](https://semver.org) notation for input so we need the full list of available versions to resolve input properly.

I would expect that the SemVer version of setup-java is parsed and converted into the version the respective distribution API expects. At least for the most common use-cases of `8` and `11` there should be no more than one request. You can always fallback to getting a wider range of versions if the action input is not sufficiently concise.

> Actually, getting all versions is not a problem. Adopt just provides about ~50 versions that is pretty small list. It takes less than 0.1 seconds to retrieve it. It is not a problem even for Azul that provides wider list of available versions.
> 
> We don't see any problems with it

Not yet.

And even if downloading the entire list is usually fast and works most of the time, an entire build would fail when just one of the paged (and unnecessery) requests fails - something that can be avoided for most use-cases.

giltene:
> ,,,Images will contain latest version of every LTS and will be updated with 1-week delay.
> First, action tries to find version that satisfies input locally. If it is not found locally, it will download and install version in runtime.
> If you specify exact version like `11.0.10+9` but image contains only `11.0.10+8`, then `11.0.10+9` will be downloaded in runtime.

@maxim-lobanov I think this will cause a lot of issues. Bringing in a later-than-cached-version only of it is explicitly specified is both counter-intuitive and will likely break many of the commonly relied on behaviors of setup-java.

For example, a common [best] practice is to run tests both against a statically and explicitly selected update version (e.g. "11.0.7") and the latest update version (e.g. "11") of the same java versions, for any version that is currently maintained and has new updates posted multiple times per year. The simple reason for this is to help triage and identify whether tests break because of a change in the jdk or a change in the repo. It is [unfortunately] common for new updates to break things. E.g. just in the recent few months, OpenJDK 8u262, 8u272, and 11.0.9 all had significant real-world breakages found post release, leading to the release of the out of cycle 8u265, 8u275, and 11.0.9.1 within weeks. Things that had such breakages revealed in post-release tests included e.g. hadoop and solo.

A one week delay in testing new updates due to caching of e.g. LTS updates would create wide ranging negative impacts, both for the tested repos (which will find out a week late that stuff is newly broken that they may need to be aware of work around, or that new out of cycle stuff has a fix that needs to be urgently used), and for OpenJDK in general: the wide ranging testing of new updates in the hours and days immediately after release can help drive the feedback that identify these things. With 60K+ Github workflows currently testing openjdk updates as they appear, this quick post-release exposure is very helpful, and delaying it by a week would be sad.

I'd like to suggest that, at the very least, the lookup and resolution of the semver specified versions to URIs be done on each run with no attempt to cache that logic, but that the download URI itself be used as a key for caching. This way updates will show up immediately in tests that run after they appear. A preferred solution would cache the hash or checksum to make sure that even replaced binaries at the same URI are treated properly, but URI based caching is probably the next best thing.

maxim-lobanov:
By default, action should always check cache firstly. The most of users prefer to use cached version (speed up their builds by 10-30 seconds and more reliable since there is no any change to fail randomly via connection issue) but get up-to-date version with a few days delay. The same way like [actions/setup-go](https://github.com/actions/setup-go), [actions/setup-python](https://github.com/actions/setup-python), [actions/setup-node](https://github.com/actions/setup-node) work.

The use-case that you have described makes sense but it shouldn't be a default behavior. For Node.JS, it is achieved via separate flag [check-latest](https://github.com/actions/setup-node/blob/main/action.yml#L12) so probably we should consider adding something similar to cover this use-case.

maxim-lobanov:
Merged these changes to `v2-preview` branch. Feel free to create an issue for any suggestions / found bugs.
The `v2-preview` version can be consumed via
```
- name: setup-java
  uses: actions/setup-java@v2-preview
  with:
    java-version: 'adopt'
    distribution: '11'
```

ibauersachs:
I have to admit, I don't know where the cache duration is handled. But I have to concur with @giltene's reasoning.

> By default, action should always check cache firstly. The most of users prefer to use cached version (speed up their builds by 10-30 seconds and more reliable since there is no any change to fail randomly via connection issue) but get up-to-date version with a few days delay.

Caching, yes, please. But definitely not for a week! If I specify `java-version: 11` I expect to get the newest release within hours of it's availability.

Also, you have discarded my comment about downloading the entire list of available JDKs and the possible connection failures. Now you're using the exact same argument for an insanely long cache timeout.
You can always fall back to using the cache (if available) on connection failures and report a warning.

> The same way like [actions/setup-go](https://github.com/actions/setup-go), [actions/setup-python](https://github.com/actions/setup-python), [actions/setup-node](https://github.com/actions/setup-node) work.
> The use-case that you have described makes sense but it shouldn't be a default behavior. For Node.JS, it is achieved via separate flag [check-latest](https://github.com/actions/setup-node/blob/main/action.yml#L12) so probably we should consider adding something similar to cover this use-case.

A flag/timeout option is fine with me, but it should default to getting new releases fast.

giltene:
Caching the contents of a chosen JDK download makes perfect sense, and can save 20-30 seconds and a bunch of bandwidth. Catching the decision of which JDK to choose to download does not. It would only save a small fraction of a second and practically no bandwidth, and lead to operational changes compared to the current behavior seen by the many users of 1.x. 

In addition, inconsistent functionality behavior that depends on under-the-hood  choice to selectively cache JDKs (only LTS? Only some distros?) in ways that cause widely varying behavior in terms of how up to date the JDK is (compared to what is currently available for that version spec and distro) will add “surprising” outcomes. Java has an industry-aligned cadence for quarterly updates (across all coordinating distros), so varying this behavior by version spec and distro will lead to some weird stuff.

is there any reason not to just reactively cache all resolved JDK URIs for some reasonable amount of time? Or maybe cache the few hundred recent or hotly used ones of the number or space is a concern?

ibauersachs:
@maxim-lobanov would you please get back to the caching issue? Just not answerig anymore isn't a solution. Or did I miss a separate thread?

maxim-lobanov:
As for the caching timeline - Hosted images are updated on weekly basis so the image cache can't be updated more often.
As for the caching behavior:
> Context: we have an image with cached Java 11.0.7+3. Yesterday, Java 11.0.7+4 was released.
User specifies task input `java-version: 11`

We have two different behavior options:
1) Java 11.0.7+3 will be chosen without delay and without checking remote source since it is in cache and satisfies user's input. User will be able to specify exact version `11.0.7+4` as input to get the new version earlier.
2) Action will check the latest version and find that 11.0.7+4 is available. Then action will download and install 11.0.7+4 in runtime with some time delay.

@konradpabjan , The first option will satisfy users who more interested in getting fast and lower-cost builds and don't care much about 1-week delay with updates. We use this approach with other tools and it works perfectly.
The both cases make sense for me but in my understanding, the default task behavior should be the 1st option to use image cache as much as possible and reduce customers' build time like other 1st-party actions do. 

In `setup-node`, we resolved this case with additional flag like `check-latest` that forces action to use the 2nd option and keeping 1st option by default.

ibauersachs:
As @giltene already mentioned, the delay of case 2) is only there if there is actually a newer version. In the majority of builds, there is no newer version and the only delay is the API call, not the actual JRE/JDK download.
If you're so concerned about API call timeouts and delays, make the behavior configurable with `check-latest` **and** reduce the amount of API calls necessary to retrieve the wanted version.

konradpabjan:
> The first option will satisfy users who more interested in getting fast and lower-cost builds and don't care much about 1-week delay with updates. We use this approach with other tools and it works perfectly.

I really like the approach that `setup-node` takes. I think most users would prefer to have their workflows just run fast and use whatever is cached on the image by default. In the scenario where a very recent version is needed that is not cached then an extra input such as `check-latest` makes sense.

@maxim-lobanov let's try to squeeze that into `v2-preview` 😃 

Thanks everyone for the feedback and discussion!

dmitry-shibanov:
Hello everyone. We prepared [pull request](https://github.com/actions/setup-java/pull/141) with adding check-latest flag. At the moment the tests are failing but it will be fixed by this [pull request](https://github.com/actions/setup-java/pull/139).

PR #134, opened 2021-03-08
actions/setup-java@v2 - Documentation

maxim-lobanov:
**Description:**
This pull-request updates documentation according to setup-java@v2

Looks like we had a ton of different use-cases in README file and it becomes unreadable. I have decided to leave multiple base use-cases in README and move other use-cases to the separate file `advanced-usage.md`.
Also, added guide for migration from V1 to V2 since it brings a bunch of breaking changes.

I would be very appreciate for some review and suggestions about wording to sound native :) 

Note: we have decided to split changes to a few pull-requests to simplify review process:
- Installers: https://github.com/actions/setup-java/pull/132
- Tests: https://github.com/actions/setup-java/pull/133
- Docs: (this pull-request)
- Related ADR: https://github.com/actions/setup-java/pull/97


README.md
```diff
@@ -4,252 +4,83 @@
...
-- registering problem matchers for error output
-
-# Usage
-
-See [action.yml](action.yml)
-
-## Basic
+This action provides the following functionality for GitHub Actions runners:
+- Downloading and setting up a requested version of Java. See [Usage](#Usage) section for the list of supported distributions
+- Extracting and caching custom version of Java from local file
```
konradpabjan, line 9: `See [Usage](#Usage) for a list of supported distributions`. No need to have the "section" after the hyperlink

`Extracting and caching custom version of Java from a local file`

README.md
```diff
@@ -4,252 +4,83 @@
...
+- Configuring runner for publishing using Apache Maven
+- Configuring runner for publishing using Gradle
+- Configuring runner for using GPG private key
+- Registering problem matchers for error output
+
+## V2 vs V1
+- V2 has support of custom distributions and provides support of Zulu OpenJDK and Adoptium (former AdoptOpenJDK) out of the box. V1 supports only Zulu OpenJDK
```
konradpabjan, line 16: I would switch (former AdoptOpenJDK) to (formerly known as AdoptOpenJDK)

README.md
```diff
@@ -4,252 +4,83 @@
...
-  e.g. ```8.0.232, 7.0.181, 11.0.4```
-
-  e.g. ```8.0.x, >11.0.3, >=13.0.1, <8.0.212```
 
-- An early access (EA) Java version
+#### Supported version syntax
+Input `java-version` supports version range or exact version in [SemVer](https://semver.org/) format:
+- major versions: `8`, `11`, `15`, `11.x`
+- more specific versions: `8.0.232`, `11.0.4`, `11.0`, `11.0.x`
+- an early access (EA) versions: `15-ea`, `15.0.0-ea`, `15.0.0-ea.2`
```
konradpabjan, line 37: `The "java-verion" input supports an exact version or a version range using [SemVer](https://semver.org/) notation:`

the last line I would change to just `early access...`

docs/switching-to-v2.md
```diff
@@ -0,0 +1,35 @@
+# Switching to V2
+## Java distribution
+The major breaking change in V2 is the new mandatory input `distribution`. This field should be specified with one of supported distributions. See [Supported distributions](../README.md#Supported-distributions) section for the list of available options.
```
konradpabjan, line 3: First sentence: `The major breaking change in V2 is the new mandatory "distribution" input.`

Also: `See [Supported distributions](../README.md#Supported-distributions) for a list of available options.`

docs/switching-to-v2.md
```diff
@@ -0,0 +1,35 @@
+# Switching to V2
+## Java distribution
+The major breaking change in V2 is the new mandatory input `distribution`. This field should be specified with one of supported distributions. See [Supported distributions](../README.md#Supported-distributions) section for the list of available options.  
+Use `zulu` keyword if you would like to continue using the same distribution as in the V1.
```
konradpabjan, line 4: Use **the** `zulu keyword

docs/switching-to-v2.md
```diff
@@ -0,0 +1,35 @@
...
+    distribution: 'zulu'
+    java-version: '11'
+    java-package: jdk # optional (jdk or jre) - defaults to jdk
+- run: java -cp java HelloWorldApp
+```
+
+**General recommendation** — configure CI with the same distribution that is used on your local dev machine.
+
+## Installing custom Java distribution from local file
+Since `distribution` input is required in V2, you should specify it as `jdkFile` to continue installing Java from local file
```
konradpabjan, line 19: Since the `distribution` input is required in V2, you should specify it using `jdfFile` to continue installing Java from a local file on the runner

docs/switching-to-v2.md
```diff
@@ -0,0 +1,35 @@
...
+- uses: actions/setup-java@v2-preview
+  with:
+    distribution: 'jdkFile'
+    jdkFile: ${{ runner.temp }}/java_package.tar.gz
+    java-version: '11.0.0'
+    architecture: x64
+```
+
+## Dropping legacy Java version syntax 1.x
+V1 supported legacy Java syntax like `1.8` (same as `8`), `1.8.0.212` (same as `8.0.212`).
```
konradpabjan, line 34: Since there are only 2 examples, it best to just use `and` so: `V1 supported legacy Java syntax such as 1.8 (same as 8) and 1.8.0.212 (same as 8.0.212).`

docs/switching-to-v2.md
```diff
@@ -0,0 +1,35 @@
...
+  with:
+    distribution: 'jdkFile'
+    jdkFile: ${{ runner.temp }}/java_package.tar.gz
+    java-version: '11.0.0'
+    architecture: x64
+```
+
+## Dropping legacy Java version syntax 1.x
+V1 supported legacy Java syntax like `1.8` (same as `8`), `1.8.0.212` (same as `8.0.212`). 
+V2 dropped support of old syntax so workflows should be changed accordingly.
```
konradpabjan, line 35: `V2 dropped support for legacy syntax so workflows should be updated accordingly`

docs/advanced-usage.md
```diff
@@ -0,0 +1,290 @@
...
+- [Installing custom Java distribution from local file](#Installing-Java-from-local-file)
+- [Testing against different Java distributions](#Testing-against-different-Java-distributions)
+- [Testing against different platforms](#Testing-against-different-platforms)
+- [Publishing using Apache Maven](#Publishing-using-Apache-Maven)
+- [Publishing using Gradle](#Publishing-using-Gradle)
+
+See [action.yml](../action.yml) for more details on task inputs.
+
+## Selecting Java distribution
+Input `distribution` is mandatory and should be provided to use action. See [Supported distributions](../README.md#Supported-distributions) for the list of available options.
```
konradpabjan, line 16: Change `and should be provided to use action` to `needs to be provided`. Also `for the list of available options` to `for a list of available options`

docs/advanced-usage.md
```diff
@@ -0,0 +1,290 @@
...
+- uses: actions/setup-java@v2-preview
+  with:
+    distribution: '<distribution>'
+    java-version: '11'
+    architecture: x86 # optional - defaults to x64
+- run: java -cp java HelloWorldApp
+```
+
+## Installing Java from local file
+If your use-case requires custom distribution or version that is not provided by action, you can download it manually and then action will take care about installation and caching of this version on VM:
```
konradpabjan, line 68: If your use-case requires a custom distribution or a version that is not provided by setup-java, you can download it manually and setup-java will take care of the installation and caching on the VM:

docs/advanced-usage.md
```diff
@@ -0,0 +1,290 @@
...
+        uses: actions/setup-java@v2-preview
+        with:
+          distribution: 'adoptium'
+          java-version: ${{ matrix.java }}
+      - run: java -cp java HelloWorldApp
+```
+
+## Publishing using Apache Maven
+
+### Extra setup for pom.xml:
```
konradpabjan, line 128: I would move this section below the `### Yaml example:` part

AlenaSviridenko (approved):

gdams (approved):

docs/advanced-usage.md
```diff
@@ -0,0 +1,289 @@
...
+- [Installing custom Java distribution from local file](#Installing-Java-from-local-file)
+- [Testing against different Java distributions](#Testing-against-different-Java-distributions)
+- [Testing against different platforms](#Testing-against-different-platforms)
+- [Publishing using Apache Maven](#Publishing-using-Apache-Maven)
+- [Publishing using Gradle](#Publishing-using-Gradle)
+
+See [action.yml](../action.yml) for more details on task inputs.
+
+## Selecting Java distribution
+Input `distribution` is mandatory and needs to be provided to use action. See [Supported distributions](../README.md#Supported-distributions) for a list of available options.
```
konradpabjan, line 16:
```suggestion
Input `distribution` is mandatory and needs to be provided. See [Supported distributions](../README.md#Supported-distributions) for a list of available options.
```

docs/switching-to-v2.md
```diff
@@ -0,0 +1,35 @@
+# Switching to V2
+## Java distribution
+The major breaking change in V2 is the new mandatory `distribution` input. This field should be specified with one of supported distributions. See [Supported distributions](../README.md#Supported-distributions) for a list of available options.  
+Use the `zulu` keyword if you would like to continue using the same distribution as in the V1.
```
konradpabjan, line 4:
```suggestion
Use the `zulu` keyword if you would like to continue using the same distribution as in V1.
```

konradpabjan (approved):

README.md
```diff
@@ -4,252 +4,83 @@
...
+- Configuring runner for publishing using Gradle
+- Configuring runner for using GPG private key
+- Registering problem matchers for error output
+
+## V2 vs V1
+- V2 supports custom distributions and provides support for Zulu OpenJDK and Adoptium (formerly known as AdoptOpenJDK) out of the box. V1 supports only Zulu OpenJDK
+- V2 requires you to specify distribution along with the version. V1 defaults to Zulu OpenJDK, only version input is required. Follow [the migration guide](docs/switching-to-v2.md) to switch from V1 to V2
+
+## Usage
+Input `distribution` is mandatory. See [Supported distributions](../README.md#Supported-distributions) section for the list of available options.
```
konradpabjan, line 20:
```suggestion
Input `distribution` is mandatory. See [Supported distributions](../README.md#Supported-distributions) for a list of available options.
```

MaksimZhukov (approved):

giltene:
The use of "adoptium"  as a distribution name that is actuality will be getting the AdoptOpenJDK distro builds seems wrong here. So are the links to the AOJ site instead of the Adoptium site.

AFAIK [Adoptium](https://projects.eclipse.org/projects/adoptium) as an Eclipse project is a "...continuation of the original AdoptOpenJDK mission", but Eclipse Adoptium builds will be Eclipse Adoptium builds, and not AdoptOpenJDK builds. Adoptium and AdoptOpenJDK are two separate distros, produced by separate organizations, using separate tests and governance, and [potentially] separate terms of use. The actual binaries will [presumably] show different distro names in their version strings. And for a given java version sepcified [e.g. a past 8.0.252, or a future 11.0.19] there may be distro binaries available from one of those two distros, or both, or neither. 

Explicitly specifying one distro in the "distribution" field and getting another distorts binaries would clearly be wrong.

So either the word "adoptium" should be changed to something that properly refers to AdoptOpenJDK distro and will not conflict with Adoptium distro when it starts providing binaries, or the meaning of using "adoptium" should change to [only] using bits provided by the Eclipse Adoptium project, and not bits that come from some other distro.

Note: If the choice is made to change the name used to specify the AdoptOpenJDK distro, that choice should likely avoid using the word "openjdk" in the distro name for the obvious reasons (including trademark rules and the avoidance of assymetric use of the term openjdk across the various choices for distros of OpenJDK). Names like "adopt" and "AOJ" probably work in this regard.

konradpabjan:
Thanks for the info @giltene 

My assumption based off of https://blog.adoptopenjdk.net/2020/06/adoptopenjdk-to-join-the-eclipse-foundation/ was that down the road `AdoptOpenJDK` would be transitioned to `Eclipse Adoptium` and things would be largely the same. Two separate distros seems weird (although it's possible of course). I think the confusion here is because the rename that was announced hasn't happened yet and it will probably take some time.

> As part of a move to the Eclipse Foundation, we will undergo a project rename.

As well as

>  In time, some services, websites, and APIs would move to new URLs, and some package names might change. Everything will be announced well in advance on a per-service basis with long transition periods.

Even if you go to https://projects.eclipse.org/projects/adoptium/downloads there is currently nothing.

It is confusing though and I like the suggestion to use `adopt`. Both users of `AdoptOpenJDK` and `Eclipse Adoptium` or `Adoptium`  should be able to recognize it. @maxim-lobanov lets go with that

giltene:
In the "basic" example description in the README, we should change the distro selection from "adopt" to "zulu". (We can also mention the other distro options explicitly in the comment, to avoid having people jump somewhere else just to find out possible spelling). For the 60K+ current users of V1, filling in zulu as the distribution (which is a new required property) will be the "compatible" entry, and is therefore the obvious example choice if one chose is shown. It is the most straightforward way to not break (especially since there are lots of currently used versions in workflow that zuliu provides and adopt does not).

gdams:
@giltene no users will be broken? All V1 users will continue to be able to use Zulu going forwards. If they are explicitly making the decision to bump to V2 then they can also make the code change to choose their vendor. As pointed out before, AdoptOpenJDK is already preinstalled on all actions environments so using adopt in the example seems the most logical approach going forwards.

konradpabjan:
@giltene We're going to update the `basic` example to have an example with Zulu as well https://github.com/actions/setup-java/pull/136/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R34-R44

Given that `adopt` is already pre-installed on our images, for consistency moving forward we'll be using `adopt` in examples so expect public documentation at some point to reflect those changes: https://docs.github.com/en/actions/guides/building-and-testing-java-with-maven

As mentioned if users are pinned to `v1` then nothing will change or break. Only users pinned to `@main` will have workflows start to fail but we have a [migration guide](https://github.com/actions/setup-java/blob/v2-preview/docs/switching-to-v2.md) and we will have a change-log the day of the `v2` release. We also actively don't recommend users pin to `@main` or `@master` (and instead to tags) as breaking changes can happen between major versions.

PR #136, opened 2021-03-16
Add "overwrite-settings" input parameter

maxim-lobanov:
**Description:**
In some scenarios there's already a valid Maven settings file which shouldn't or cannot be overwritten by actions/setup-java.

The `overwrite-settings` (default: `true`) configuration setting enables users to disable the generation of the Maven settings file by actions/setup-java.

This pull-request was inspired by https://github.com/actions/setup-java/pull/82, just adopt changes to V2-preview branch.

Included changes:
- Add `overwrite-settings` input
- Add e2e and unit tests to cover new logic
- Implement `getBooleanInput` util func based on https://github.com/actions/checkout/blob/main/src/input-helper.ts#L118
- Move logic to read settingsDirectory to `configureAuthentication` function to have all related logic at the same place.
- Update README to include usage example for both Zulu and Adopt

**Related issue:**
Closes https://github.com/actions/setup-java/issues/79

**Check list:**

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -114,7 +114,7 @@ async function write(directory: string, settings: string, overwriteSettings: boo
     core.info(`Writing to ${location}`);
   } else {
     core.info(
-      `Skipping generation '${location}' - it already exists and overwriting is not required`
+      `Skipping generation ${location} because file already exists and overwriting is not required`
     );
     return;
   }
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -8,6 +8,7 @@ import * as os from 'os';
 import { create as xmlCreate } from 'xmlbuilder2';
 import * as constants from './constants';
 import * as gpg from './gpg';
+import { getBooleanInput } from './util';
 
 export const M2_DIR = '.m2';
 export const SETTINGS_FILE = 'settings.xml';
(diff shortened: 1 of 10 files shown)


maxim-lobanov:
cc: @MaksimZhukov @dmitry-shibanov @vsafonkin  @AlenaSviridenko to review this pull-request

dmitry-shibanov (approved):

src/auth.ts
```diff
@@ -16,6 +17,10 @@ export async function configureAuthentication() {
   const id = core.getInput(constants.INPUT_SERVER_ID);
   const username = core.getInput(constants.INPUT_SERVER_USERNAME);
   const password = core.getInput(constants.INPUT_SERVER_PASSWORD);
+  const settingsDirectory =
+    core.getInput(constants.INPUT_SETTINGS_PATH) || path.join(os.homedir(), M2_DIR);
+  // Consider setting overwriteSettings = false by default in next major release
+  const overwriteSettings = getBooleanInput(constants.INPUT_OVERWRITE_SETTINGS, true);
```
vsafonkin, line 23: As I see, it always returns `true` if the second parameter is `true`, for any inputs, is it expected? or I miss something?

vsafonkin (approved):

MaksimZhukov (approved):

src/auth.ts
```diff
@@ -16,6 +17,10 @@ export async function configureAuthentication() {
   const id = core.getInput(constants.INPUT_SERVER_ID);
   const username = core.getInput(constants.INPUT_SERVER_USERNAME);
   const password = core.getInput(constants.INPUT_SERVER_PASSWORD);
+  const settingsDirectory =
+    core.getInput(constants.INPUT_SETTINGS_PATH) || path.join(os.homedir(), M2_DIR);
+  // Consider setting overwriteSettings = false by default in next major release
```
konradpabjan, line 22: This comment can be removed given that we're keeping it to `true` for this release

konradpabjan (approved):

src/auth.ts
maxim-lobanov: I am mostly added it for V3 :) but I can remove

konradpabjan: `v3` might be a long long time out so that's why I'm inclined to just remove it now 😄 Setting it to `false` in the future would also break the easy setup we have with GitHub package registry so I don't think down the line that will be an easy choice that is made

maxim-lobanov: Removed

PR #139, opened 2021-03-19
Add "Contents/Home" postfix on macOS if provider creates it

maxim-lobanov:
**Description:**
JDK folder on MacOS can contain additional nested folders for some providers.
Previously, we have included these folders to PATH only when download Java from remote. We should do the same when resolve Java from tool-cache. So decide to move this logic to base class to avoid duplication.

image

### DIFF
--- a/.github/workflows/e2e-versions.yml
+++ b/.github/workflows/e2e-versions.yml
@@ -33,6 +33,15 @@ jobs:
       - name: Verify Java
         run: bash __tests__/verify-java.sh "${{ matrix.version }}" "${{ steps.setup-java.outputs.path }}"
         shell: bash
+       - name: setup-java
+        uses: ./
+        id: setup-java2
+        with:
+          java-version: ${{ matrix.version }}
+          distribution: ${{ matrix.distribution }}
+      - name: Verify Java
+        run: bash __tests__/verify-java.sh "${{ matrix.version }}" "${{ steps.setup-java2.outputs.path }}"
+        shell: bash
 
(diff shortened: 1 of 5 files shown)


konradpabjan (approved):

sormuras:
Merging this PR could render my [repackage work-around](https://github.com/sormuras/download-jdk/blob/09f98095e4e40abd92c9d6ffc0744c73bb9166d9/action.yml#L32-L42) I perform at https://github.com/sormuras/download-jdk action...

```yml
    - name: 'Repackage tarball on Mac OS'
      shell: bash
      run: |
        if [ "$RUNNER_OS" == "macOS" ]; then
          echo 'Repackage tarball on Mac OS'
          JDK_EXTRACTED=${HOME}/downloaded-jdk-${{ inputs.feature }}
          mkdir ${JDK_EXTRACTED}
          tar --extract --file ${JDK_FILE} -C ${JDK_EXTRACTED} --strip-components 4
          rm ${JDK_FILE}
          tar --create --file ${JDK_FILE} -C ${HOME} downloaded-jdk-${{ inputs.feature }}
        fi
```

...superseded.

dmitry-shibanov (approved):

sormuras:
Indeed, after upgrading to `setup-java@v2`, I could remove the work-around via https://github.com/sormuras/download-jdk/commit/31721712feafec2b584cb06298855cf98e050ebd

Any chance to backport this feature to `v1`?

PR #145, opened 2021-03-25
Avoid "+" sign in Java path in v2-preview

maxim-lobanov:
**Description:**
Initially, we have implemented logic to save Java on local disk using Java version semver notation: `/hostedtoolcache/Java/11.0.3+4`.

Recently, we have faced with the issue on images side that it breaks some tricky Kotlin and Java use-cases. Find more details in https://github.com/actions/virtual-environments/issues/3014

As a fix, we are replacing `+` to `-` during reading versions from local cache and when put them to local cache.
So it will be store Java on path like `/hostedtoolcache/Java/11.0.3-4`. It also works for Adopt versions like `11.0.3+4.2.256`, it will be `11.0.3-4.2.256`
As a additional benefit: with this approach toolkit/toolcache is able to resolve 4-digits versions in tool-cache properly
Also added additional unit-tests to cover it


dmitry-shibanov (approved):

konradpabjan:
One small issue, I'm looking at the E2E tests and the output shows that the path has a `+`, I think that should be different.

Here is a output for Adopt

image

Here is Zulu

image

konradpabjan (approved):

Issue #142, opened 2021-03-19
Add support for multiple distributions via the Disco API


giltene:
Description:
Add support for a new, optional "distro" property, and use the [Foojay Disco API](https://github.com/foojay2020/discoapi) to locate JDK/JRE binaries for the requested distro, such that any distro supported by Disco (now or in the future) will automatically be enabled in actions/setup-java with no breaking changes to existing workflows needed.

Justification:
The ability to specify an OpenJDK distribution other than the default is often requested, and can be reflected in e.g. https://github.com/actions/setup-java/issues/13,  https://github.com/actions/setup-java/issues/68, https://github.com/actions/setup-java/issues/69, https://github.com/actions/setup-java/issues/70, https://github.com/actions/setup-java/issues/71, https://github.com/actions/setup-java/issues/72 . 

The [Foojay Disco  API](https://github.com/foojay2020/discoapi) was formed to provide a common, level-playing-field, community driven means of locating OpenJDK distribution binaries. It is specifically formulated to support semver-based version specification (in a way that is compatible with how java-version is specified in actions/setup-java) and provide a normalized way to translate a specification of a Java version spec and java package type to a URI pointing to matching OpenJDK distro binary if one exists. Using Disco will greatly simplify setup-java logic, and remove significant work (scanning CDNs or paginating through per-distro APIs in search of sever-satisfying matches in lists of possible binaries). It will also allow any new distros (once knowledge about them is added to Disco) to be immediately used in actions/setup-java with no change to the action code itself. The Disco project is committed to providing good coverage of OpenJDK and Java distros, and is already covering 9+ existing OpenJDK distributions in a way that is ready to tap here.

While some incremental work towards supporting the multitude of OpenJDK distributions has been contemplated and prototyped in v2.x, there are still some significant controversies around the form that support will take: E.g. will it be a breaking change, or an added [optional] property such that existing v1.x workflows will be able to move to v2.x without syntax modification or serious behavioral change?

Adding an optional property to v1.x, which will immediately and directly address (https://github.com/actions/setup-java/issues/13,  https://github.com/actions/setup-java/issues/68, https://github.com/actions/setup-java/issues/69, https://github.com/actions/setup-java/issues/70, https://github.com/actions/setup-java/issues/71, https://github.com/actions/setup-java/issues/72),  without creating a braking change or incompatibility to any of the 60K+ workflows using actions/setup-java right now, is quite straightforward. I propose that we move ahead with that while the v2.x stuff gets worked out in time.

[...]

Issue #137, opened 2021-03-16
Breaking changes in "main" branch on 5 April


maxim-lobanov:
Hello everyone!

Recently, we have published [actions/setup-java@v2-preview](https://github.com/actions/setup-java/tree/v2-preview) with support of Zulu and Adopt distributions.
We are going to make the stable release and push `v2-preview` changes on `main` on 5 April 2021.

It won't affect customers who pin on specific tag of action like `actions/setup-java@v1` or `actions/setup-java@1.1.0` but it will break customers who pin their workflows to `actions/setup-java@main`. Workflows will be broken because V2 has required input `distribution`. According to [GitHub Actions documentation](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-branches), it is not recommended to pin workflow to main branch.

To mitigate the issue, we suggest pinning your workflow to `actions/setup-java@v1`.

giltene:
The proposed timeframe (for pushing breaking changes to main) is way too early IMO, unless one intends to change behavior from this point forward, and to keep adding breaking changes to the “main” branch every few weeks from now on, prevention “main” from stabilizing in a non-breaking-change manner.

Given the breaking nature of multiple new behaviors proposed in the current v2 preview, and the lack (this far) of a publicly accessible description of the actual set of v2 behavior changes intended, significantly more user feedback (as opposed to action developer feedback) on the v2 preview version is needed, with time to react to that feedback needed and potential break-the-breaking-changes needed as well.

Some proposed behavioral changes that I am aware of so far include a complete change to the means by which distro binaries are pulled (cdn scraping to api), newly introduced (and very significant) update lag times, new required properties, and new properties with defaults that change behavior quite significantly (and sometimes subtly but materially) from what a v1.x setup-java does today. The dust has not settled on most of those, and user feedback on their impact is still pending. The preview v2-branch is a fine place to continue gathering that feedback and potentially adjusting choices in ways that would “break” v2-preview to v2-preview compatibility. It would be preferable IMO for v2 preview to reach an empirically stable level, where no breaking changes have been done within it for a while, before merging into the “main” branch.

This, obviously, goes doubly so for v2 itself. Since the next opportunity to create any breaking changes (that will break with those that will actually get included in the first non-preview v2) will be a long time down the road, in a v3,, it would be advisable to avoid rushing in a set of breaking changes without allowing the v2 preview to bake in actual user experience, feedback and refinement for a sufficient amount of time to gauge stability.

[...]

PR #147, opened 2021-03-31
Enhance v1 with the ability to select a distribution

HanSolo:
**Description:**
Instead of using only the Zulu distribution, this PR makes use of the [foojay discoapi](https://github.com/foojay2020/discoapi) to select from different distributions (supported are AdoptOpenJDK, Corretto, Dragonwell, Liberica, OJDK Build, Oracle OpenJDK, SAP Machine, Trava and Zulu). To choose a distribution you can use the parameter "distro" to specify the distribution. If no distribution is provided it will fall back to Zulu which is the default at the current v1.
e.g. 
```json
steps:
- uses: actions/checkout@v2
- uses: actions/setup-java@v1
  with:
    java-version: '15.0.1' # The JDK version to make available on the path.
    java-package: jdk # (jre, jdk, or jdk+fx) - defaults to jdk
    architecture: x64 # (x64 or x86) - defaults to x64
    distro: aoj # The distribution you would like to use
- run: java -cp java HelloWorldApp
```
We run a [Disco Testing Matrix](https://github.com/foojay2020/discoTestingMatrix) that checks if the specified distributions are available via the Disco API to make sure it works as expected.

**Related issue:**
[Add support for multiple distributions via the Disco API](https://github.com/actions/setup-java/issues/142)

**Check list:**


konradpabjan:
Thanks for putting out this PR @HanSolo 

I'll be frank, adding support for all these distributions would be amazing and users would love it ❤️ However we can't currently use the Disco API. 

My comment in the linked issue touches on this: https://github.com/actions/setup-java/issues/142#issuecomment-804142915

>I do have have some security concerns too that might make this a blocker (following up on this). Using the Disco API introduces a single point of failure that if compromised could for example lead to malicious binaries being downloaded. Given that the Disco API is outside of the control of GitHub and the attack surface this could be very concerning for a lot of our customers (Enterprise in particular). The API is also fairly new (~3 months).

The main concern from GitHubs perspective is security. We're releasing `v2` on Monday April 5th and that only supports Zulu and Adopt distributions by directly using the APIs provided by those distributions. We don't want to introduce a dependency or single point of failure which we absolutely do not need to. We currently do not want to use anything else which is why we explicitly opted for initially supporting just 2 distributions (the biggest ask from users that we've seen is also for `adopt`). It's more tedious and there is more maintenance to use the APIs directly, but that's what we've decided to currently use for the sake of security.

I few other things:
- `v1` is effectively in maintenance mode and only security fixes or crucial fixes will be back-ported
-  With `v2` we're introducing a breaking change with the new `distribution` input. That will now be required while with these changes there would a level of favoritism towards a single distribution which we're trying to get away from

The work in this PR is amazing and I'm wouldn't want it to go to waste (I'm sure it took a bit of time) so I highly recommend advertising your fork in some of the issues that users are requesting support for more distributions.

giltene:
> ...We're releasing `v2` on Monday April 5th ...

Separate from the valid discussion of security (for which I think Disco can provide ample answers and address as needed as part of evolving this PR), I am surprised at the notion of releasing v2 on April 5th. I've seen no discussion of such an near-term schedule for what will enshrine a set of new breaking changes (that will be hard to "break" again if they aren't quite baked yet). Moving 'main' over to v2-preview was discussed in https://github.com/actions/setup-java/issues/137 (where I separately noted that it seems to early to do that) but that is a far cry from actually releasing v2...

v2 needs more time to bake. More user feedback on v2-preview, based on actual attempted uses by actual repos, is a big part of that, and v2-preview has not been available long enough to solicit such feedback. For example, there has not yet been an OpenJDK quarterly update cycle that would reveal the impact of some of the issues being debated about some of its proposed behaviors.

A variant of this PR base don v2-preview would probably be a good way to discuss the actual support of multiple distros (which was the main goal stated for the v2 ADR).

Since this is not a breaking change to v1, and it addresses a set of popularly requested distro support, I see no reason not to consider this for v1 as well.

giltene:
The security questions are valid ones (e.g. why would one trust the distro binaries to which the disco API points to?), but it is worth noting using the Disco API does not introduce any more points-of-depedency that are outside of GitHub or setup-java control than already exist in setup-java@v1 or in the proposed setup-java@v2. E.g. the same questions can [and should?] be asked about the binaries currently being pulled from CDNs, or expected too be puled from the API-located URI endpoints in v2-preview. The proper answer to such trust questions is invariably "check the signature for the distro binary you got against the actual binary, and verify that it was signed by a trusted source". 

E.g. Zulu CDNs already provide sig files for binary bundles for this very purpose (those sig files are not currently used by setup-java@v1 to validate downloaded distro binaries, but we can certainly add verification as an enhancement). Other distros (e.g. Corretto) do so as well. If setup-java (or it's users) are concerned about the authenticity of JDK distro binaries and want to ensure they were not tamperted with, and that they were not somehow redirected to use a malicious distro binary, the proper way to overcome that concern is to validate the binaries with the signatures provided, and to validate the the signatures themselves are provided by a trusted source. There are well established ways to do that.

To address this common concern in a distro-neutral way, we are planning to add signature metadata support in the Disco API in order to deliver signatures where available. Signatures are, unfortunately, not universally available for OpenJDK distros, but we would encourage all distros to create and provide signatures for their binary artifacts in order to allow clients to establish trust in the delivered binaries. This would eliminate the Disco as a supply-chain security concern. It would then, at most, be a denial-of-service concern, but since that can be said for any and all of the individual distro CDNs and API endpoints (all of which are outside of GitHub's control, and all of which could be similarly exposed to DOS attacks), that level of concern should not block things.  

Clients like actions/setup-java, which would be able to locate the distro binary URIs *and* the associated signatures where available, can then e.g. choose to maintain a list of "approved" sources (the known signing entities for the distros they are willing to use) and only accept distro binaries signed by one of those approved entities. Clients can also choose to "take a risk" on distros that are not signed if they wish. E.g. if setup-java wants to "still" use some distros that do not currently provide signatures that can be authenticated as being created by a trusted source, implicitly trusting or hoping that the mechanism for locating binaries of those distro and the binaries themselves would not be tampered with by someone other than the actual intended publisher of those binaries, it could do that on a distro-by-distro basis.

It' is obviously worth noting that the security concern raised here about the disco API applies equally to the Zulu API and the AOJ API used in the current v2-preview, and to some degree to the Zulu CDN currently used in v1. To the degree that security and supply chain  concerns around this new dependency on a 3rd party API implementation is blocking consideration of Disco as a distro-binary locating API, it should also block v2 consideration of those other APIs. The proper solution for such concern is likely the same in all cases (establish trust in the delivered binaries, so that you don't have to worry about the api that pointed you to them being compromised).

konradpabjan:
The preview time is roughly on par with what we've done with other first party actions. We also have telemetry that gives us information about how many workflows are using the preview and we feel very confident with what we've seen. The main breaking change is the new `distribution` input but anyone pinned to `@main` right now has a warning annotation always show up. We haven't noticed any un-planned regressions or failures which would warrant us extending the preview period which is why we're sticking with April 5th as the release date.

https://github.com/actions/setup-java/blob/4e6af5a2d7d7c1d4da9ee0ba337d9100d60e0bf0/src/setup-java.ts#L60

> Since this is not a breaking change to v1, and it addresses a set of popularly requested distro support

It's not a breaking change in the sense that your workflow will fail, but it's a "breaking" change in terms of the behavior of the action. Such large changes warrant a separate release.

The biggest request has been to support `adopt` distributions. It's the oldest open issue in this repo and most commented on. It's what we've been focusing on as described here:
https://github.com/actions/setup-java/blob/main/docs/adrs/0000-v2-setup-java.md#goals--anti-goals

> The main focus of the v2 version of setup-java will be to add support for adoptium builds of openJDK in addition to Zulu builds of openJDK

All of a sudden trying to add support for more distributions right before the planned release date is a bit of a stretch that certainly wasn't a goal at least for the initial v2 release.

----

With regards to the contents of this PR and supporting the Disco API, here is something we could explore

Initial `v2` release (releasing April 5th)
- Download Zulu builds using native APIs
- Download Adoptium builds using native APIs

=> A future v2 version could do something like

- Download Zulu builds using native APIs
- Download Adoptium builds using native APIs
- Download and support a few more different distributions using the Disco API. Maybe Corretto and something else to start off with

Gradually expanding support for more distributions, not everything all at once. Down the road who knows, maybe all downloads can be moved over to Disco API or some other API depending on how it goes but the main thing is to do this in controllable, bit-sized phases that are easily reviewable. The contents of this PR can be used & adapted to get something working towards initially supporting a few extra different distributions.

----

Thanks for information about sig files, I was not aware of such as feature. Validation on the `setup-java` side would be a good thing to add and it would certainly alleviate some of our concerns. It would have been good to mention this beforehand in  #142. Large discussions like this should not be happening in PRs but in issues.

> It' is obviously worth noting that the security concern raised here about the disco API applies equally to the Zulu API and the AOJ API

Downloading malicious binaries from Zulu or AOJ is a perfectly valid concern, but adding a discoverability API in the middle increases the possible attack surface which is why we're at least hesitant about initially jumping on board and using one

maxim-lobanov:
Closing the PR because, as it was mentioned above, we don't accept any new changes to v1

PR #155, opened 2021-04-05
Added 'adopt-hotspot' and 'adopt-openj9'

MarcelCoding:
**Description:**
Added the ability to select the implementation of the adoptopenjdk distributions.

**Related issue:**
https://github.com/actions/setup-java/issues/13#issuecomment-813344524

**Check list:**

### DIFF
--- a/.github/workflows/e2e-versions.yml
+++ b/.github/workflows/e2e-versions.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        distribution: ['adopt', 'zulu']
+        distribution: ['adopt', 'adopt-openj9', 'zulu'] # internally 'adopt-hotspot' is the same as 'adopt'
         version: ['8', '11', '15']
     steps:
       - name: Checkout
@@ -91,7 +91,7 @@ jobs:
       - name: Verify Java
         run: bash __tests__/verify-java.sh "11" "${{ steps.setup-java.outputs.path }}"
         shell: bash
-  
+
   setup-java-ea-versions-zulu:
     name: zulu ${{ matrix.version }} (jdk-x64) - ${{ matrix.os }}
     needs: setup-java-major-minor-versions
--- a/README.md
+++ b/README.md
@@ -53,7 +53,8 @@ Currently, the following distributions are supported:
 | Keyword | Distribution | Official site | License |
 |-|-|-|-|
 | `zulu` | Zulu OpenJDK | [Link](https://www.azul.com/downloads/zulu-community/?package=jdk) | [Link](https://www.azul.com/products/zulu-and-zulu-enterprise/zulu-terms-of-use/) |
-| `adopt` | Adopt OpenJDK | [Link](https://adoptopenjdk.net/) | [Link](https://adoptopenjdk.net/about.html)
+| `adopt` or `adopt-hotspot` | Adopt OpenJDK Hotspot | [Link](https://adoptopenjdk.net/) | [Link](https://adoptopenjdk.net/about.html)
+| `adopt-openj9` | Adopt OpenJDK OpenJ9 | [Link](https://adoptopenjdk.net/) | [Link](https://adoptopenjdk.net/about.html)
 
 **NOTE:** The different distributors can provide discrepant list of available versions / supported configurations. Please refer to the official documentation to see the list of supported versions.
 
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -22,7 +22,7 @@ steps:
 - uses: actions/checkout@v2
 - uses: actions/setup-java@v2
   with:
-    distribution: 'adopt'
+    distribution: 'adopt-hotspot' # See 'Supported distributions' for available options @ README.md
     java-version: '11'
 - run: java -cp java HelloWorldApp
 ```
--- a/src/distributions/adopt/installer.ts
+++ b/src/distributions/adopt/installer.ts
@@ -16,11 +16,11 @@ export enum AdoptImplementation {
 }
 
 export class AdoptDistribution extends JavaBase {
-  private jvmImpl: string;
-
-  constructor(installerOptions: JavaInstallerOptions, jvmImpl: AdoptImplementation) {
+  constructor(
+    installerOptions: JavaInstallerOptions,
+    private readonly jvmImpl: AdoptImplementation
+  ) {
     super(`Adopt-${jvmImpl}`, installerOptions);
-    this.jvmImpl = jvmImpl.toLowerCase();
   }
 
   protected async findPackageForDownload(version: string): Promise<JavaDownloadRelease> {
@@ -105,7 +105,7 @@ export class AdoptDistribution extends JavaBase {
       `architecture=${arch}`,
       `image_type=${imageType}`,
       `release_type=${releaseType}`,
-      `jvm_impl=${this.jvmImpl}`
+      `jvm_impl=${this.jvmImpl.toLowerCase()}`
     ].join('&');
 
     // need to iterate through all pages to retrieve the list of all versions
--- a/src/distributions/adopt/installer.ts
+++ b/src/distributions/adopt/installer.ts
@@ -7,13 +7,20 @@ import semver from 'semver';
 
 import { JavaBase } from '../base-installer';
 import { IAdoptAvailableVersions } from './models';
-import { JavaInstallerOptions, JavaDownloadRelease, JavaInstallerResults } from '../base-models';
-import { MACOS_JAVA_CONTENT_POSTFIX } from '../../constants';
+import { JavaDownloadRelease, JavaInstallerOptions, JavaInstallerResults } from '../base-models';
 import { extractJdkFile, getDownloadArchiveExtension, isVersionSatisfies } from '../../util';
 
+export enum AdoptImplementation {
+  Hotspot = 'Hotspot',
+  OpenJ9 = 'OpenJ9'
+}
+
 export class AdoptDistribution extends JavaBase {
-  constructor(installerOptions: JavaInstallerOptions) {
-    super('Adopt', installerOptions);
+  private jvmImpl: string;
+
+  constructor(installerOptions: JavaInstallerOptions, jvmImpl: AdoptImplementation) {
+    super(`Adopt-${jvmImpl}`, installerOptions);
+    this.jvmImpl = jvmImpl.toLowerCase();
   }
 
   protected async findPackageForDownload(version: string): Promise<JavaDownloadRelease> {
@@ -70,6 +77,15 @@ export class AdoptDistribution extends JavaBase {
     return { version: javaRelease.version, path: javaPath };
   }
 
+  protected get toolcacheFolderName(): string {
+    if (this.jvmImpl === AdoptImplementation.Hotspot) {
+      // exclude Hotspot postfix from distribution name because Hosted runners have pre-cached Adopt OpenJDK under "Java_Adopt_jdk"
+      // for more information see: https://github.com/actions/setup-java/pull/155#discussion_r610451063
+      return `Java_Adopt_${this.packageType}`;
+    }
+    return super.toolcacheFolderName;
+  }
+
   private async getAvailableVersions(): Promise<IAdoptAvailableVersions[]> {
     const platform = this.getPlatformOption();
     const arch = this.architecture;
@@ -83,13 +99,13 @@ export class AdoptDistribution extends JavaBase {
       `project=jdk`,
       'vendor=adoptopenjdk',
       `heap_size=normal`,
-      `jvm_impl=hotspot`,
       'sort_method=DEFAULT',
       'sort_order=DESC',
       `os=${platform}`,
       `architecture=${arch}`,
       `image_type=${imageType}`,
-      `release_type=${releaseType}`
+      `release_type=${releaseType}`,
+      `jvm_impl=${this.jvmImpl}`
     ].join('&');
 
     // need to iterate through all pages to retrieve the list of all versions
--- a/__tests__/distributors/adopt-installer.test.ts
+++ b/__tests__/distributors/adopt-installer.test.ts
@@ -1,8 +1,6 @@
 import { HttpClient } from '@actions/http-client';
 
-import * as semver from 'semver';
-
-import { AdoptDistribution } from '../../src/distributions/adopt/installer';
+import { AdoptDistribution, AdoptImplementation } from '../../src/distributions/adopt/installer';
 import { JavaInstallerOptions } from '../../src/distributions/base-models';
 
 let manifestData = require('../data/adopt.json') as [];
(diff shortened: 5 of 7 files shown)


.github/workflows/e2e-versions.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        distribution: ['adopt', 'zulu']
+        distribution: ['adopt', 'adopt-hotspot', 'adopt-openj9', 'zulu']
```
maxim-lobanov, line 21: Considering the fact that downloading / installation logic is pretty similar for both `adopt-hotspot` and `adopt-openj9`, I don't think that we should include all these use-cases to matrix because it will produce the huge test matrix for thousands of builds that will take much time for running

I suggest
- Add only `adopt-openj9` to `setup-java-major-versions` job. So the final list will be `['adopt', 'adopt-openj9', 'zulu']`
- Don't add anything to other jobs

README.md
```diff
@@ -31,6 +31,28 @@ steps:
...
+```yaml
+steps:
+- uses: actions/checkout@v2
+- uses: actions/setup-java@v2
+  with:
+    distribution: 'adopt-openj9' # See 'Supported distributions' for available options
+    java-version: '11'
+- run: java -cp java HelloWorldApp
+```
+
```
maxim-lobanov, line 55: Considering the fact that the single difference between these examples and `adopt` is `distribution` field, I think that it should be enough to add new options only to the table Supported distributions below.

README.md
```diff
@@ -53,7 +75,9 @@ Currently, the following distributions are supported:
 | Keyword | Distribution | Official site | License |
 |-|-|-|-|
 | `zulu` | Zulu OpenJDK | [Link](https://www.azul.com/downloads/zulu-community/?package=jdk) | [Link](https://www.azul.com/products/zulu-and-zulu-enterprise/zulu-terms-of-use/) |
-| `adopt` | Adopt OpenJDK | [Link](https://adoptopenjdk.net/) | [Link](https://adoptopenjdk.net/about.html)
+| `adopt` | Adopt OpenJDK Hotspot | [Link](https://adoptopenjdk.net/) | [Link](https://adoptopenjdk.net/about.html)
+| `adopt-hotspot` | Adopt OpenJDK Hotspot | [Link](https://adoptopenjdk.net/) | [Link](https://adoptopenjdk.net/about.html)
+| `adopt-openj9` | Adopt OpenJDK OpenJ9 | [Link](https://adoptopenjdk.net/) | [Link](https://adoptopenjdk.net/about.html)
```
maxim-lobanov, line 80:
```suggestion
| `adopt` or `adopt-hotspot` | Adopt OpenJDK Hotspot | [Link](https://adoptopenjdk.net/) | [Link](https://adoptopenjdk.net/about.html)
| `adopt-openj9` | Adopt OpenJDK OpenJ9 | [Link](https://adoptopenjdk.net/) | [Link](https://adoptopenjdk.net/about.html)
```

README.md
maxim-lobanov: The same for `advanced-usage.md` page.

src/distributions/adopt/installer.ts
```diff
@@ -8,12 +8,14 @@ import semver from 'semver';
...
-import { MACOS_JAVA_CONTENT_POSTFIX } from '../../constants';
 import { extractJdkFile, getDownloadArchiveExtension, isVersionSatisfies } from '../../util';
 
-export class AdoptDistribution extends JavaBase {
-  constructor(installerOptions: JavaInstallerOptions) {
-    super('Adopt', installerOptions);
+export abstract class AdoptDistribution extends JavaBase {
+  private impl: string;
+
+  constructor(impl: string, installerOptions: JavaInstallerOptions) {
```
maxim-lobanov, line 16: Considering the fact that both Hotspot and OpenJ9 have almost identical logic, I think we don't need to create the separate classes and inheritance here.
It is enough to add `jvm-impl` parameter to `AdoptDistribution` constructor.

So in `distribution-factory.ts` we will create an object like
```
case JavaDistribution.JdkFile:
      return new LocalDistribution(installerOptions, jdkFile);
    case JavaDistribution.Adopt:
    case JavaDistribution.AdoptHotspot:
      return new AdoptDistribution(installerOptions, 'Hotspot');
    case JavaDistribution.AdoptOpenJ9:
      return new AdoptDistribution(installerOptions, 'OpenJ9');
    case JavaDistribution.Zulu:
      return new ZuluDistribution(installerOptions);
```

It will simplify code and we won't need to duplicate tests

maxim-lobanov (changes requested):
@MarcelCoding Thank you for submitting this great PR!
I have left a few comments. Please take a look.

MarcelCoding:
@maxim-lobanov thanks for your feedback. I hope I changed it as you expected.

src/distributions/adopt/installer.ts
```diff
@@ -7,13 +7,20 @@ import semver from 'semver';
...
+}
+
 export class AdoptDistribution extends JavaBase {
-  constructor(installerOptions: JavaInstallerOptions) {
-    super('Adopt', installerOptions);
+  private impl: string;
+
+  constructor(installerOptions: JavaInstallerOptions, impl: AdoptImplementation) {
+    super(`Adopt-${impl}`, installerOptions);
+    this.impl = impl.toLowerCase();
```
maxim-lobanov, line 23: Could you please rename `impl` to `jvm_imp` here and in arguments?

maxim-lobanov:
@MarcelCoding , Yes! It is what I expected. I have left minor nitpick with variable naming.
In general, looks good to me.

MarcelCoding:
I did it in camel case.

dmitry-shibanov (approved):

maxim-lobanov:
> I did it in camel case.

Camel case is definitely more preferable :) 
Looks like you didn't run `npm run release` with latest commit

@konradpabjan could you please review this PR too?

maxim-lobanov (approved):

MarcelCoding:
Sorry, I did it now.

konradpabjan (approved):
Looks great! Thanks for the contribution ❤️ 

As pointed out earlier in some comments by @maxim-lobanov, the test matrix is getting a little bit large but that's up to us to sort out. If we add more distributions then we could easily have 150+ or 200 builds per PR and that's getting into overkill territory that can start to cause delays for other PRs. We need some way better way to control which tests kick off per type of PR

giltene:
Before adding additional distro names like this incrementally and/or choosing their names on an ad-hoc basis, it would be useful to examine the full set of names expected for the distribution property, thinking a few months and steps ahead. The names chosen will potentially end up being “set in stone”, so thinking of what happens to them e.g. 1 or 2 years down the road is important.

Other (currently available) distributions and possible names to be added include e.g. corretto, liberica, sapmachine, oracle-openjdk. 

We should also consider the much anticipated appearance of Eclipse Adoptium related distributions, and the likelihood of multiple distributions being referred to from [Eclipse Adoptium](https://adoptium.net/), using the Eclipse Adoptium build or quality mechanisms and/or infrastructure. One (but not the only one) of those Adoptium related those distributions will be [Eclipse Temurin](https://projects.eclipse.org/projects/adoptium.temurin) which will likely, over time, replace the current adopt-hotspot name proposed here, but will probably be properly named and referred to as “temurin” or “eclipse-temurin” rather than “adopt” or “adopt-hotspot”. Additional distributions are expected there, including e.g. the newly announced Microsoft build of OpenJDK. (So “microsoft”? “microsoft-
adoptium”? “adoptium-microsoft”?). There may also (likely?) be something else, with a different name, that replaces the currently proposed adopt-openj9 (as AdoptOpenJDK itself transitions away and is displaced by Eclipse Temurin) which may be built and tested using Adoptium project materials, but will be unlikely to be using the Eclipse Temurin name.

docs/advanced-usage.md
```diff
@@ -22,7 +22,7 @@ steps:
 - uses: actions/checkout@v2
 - uses: actions/setup-java@v2
   with:
-    distribution: 'adopt'
+    distribution: 'adopt' # See 'Supported distributions' for available options @ README.md
```
bmarwell, line 25: Why not change it to `adopt-hotspot` to [make it less ambiguous](https://github.com/actions/setup-java/issues/160)?

bmarwell (changes requested):
Just a nit (thanks @dmitry-shibanov for pointing me here).
Otherwise you can also add `fixes: #160` to the PR message now :)

src/distributions/adopt/installer.ts
```diff
@@ -7,13 +7,20 @@ import semver from 'semver';
...
+  OpenJ9 = 'OpenJ9'
+}
+
 export class AdoptDistribution extends JavaBase {
-  constructor(installerOptions: JavaInstallerOptions) {
-    super('Adopt', installerOptions);
+  private jvmImpl: string;
+
+  constructor(installerOptions: JavaInstallerOptions, jvmImpl: AdoptImplementation) {
+    super(`Adopt-${jvmImpl}`, installerOptions);
```
maxim-lobanov, line 22: There is a one problem with these changes. After these changes, Java will be located into `Java_Adopt-Hotspot_jdk` and `Java_Adopt-OpenJ9_jdk` folders in toolcache. See https://github.com/actions/setup-java/blob/main/src/distributions/base-installer.ts#L66

The problem is that Hosted images already have Adopt HotSpot cached on images in folder `Java_Adopt_jdk` so after these changes, cached versions won't be found.

I think we can fix it by overriding `toolcacheFolderName` in `AdoptDistribution` class.
Something like that:
```ts
protected get toolcacheFolderName(): string {
    if (this.jvmImpl === AdoptImplementation.Hotspot) {
      // exclude Hotspot postfix from distribution name because Hosted runners have pre-cached Adopt OpenJDK under "Java_Adopt_jdk"
      return `Java_Adopt_${this.packageType}`;
    }
    return super.toolcacheFolderName()
}
```

docs/advanced-usage.md
maxim-lobanov: I agree, that makes sense

maxim-lobanov:
@MarcelCoding I have left one more comment. We need small trick for HotSpot to make sure that pre-cached versions will be consumed on Hosted runners.

MarcelCoding:
@maxim-lobanov Ok, I fixe it.

MarcelCoding:
and @bmarwell

src/distributions/adopt/installer.ts
```diff
@@ -70,6 +77,15 @@ export class AdoptDistribution extends JavaBase {
     return { version: javaRelease.version, path: javaPath };
   }
 
+  protected get toolcacheFolderName(): string {
+    if (this.jvmImpl === AdoptImplementation.Hotspot) {
```
maxim-lobanov, line 81: Should be `AdoptImplementation.Hotspot.toLowerCase()`
Sorry for providing invalid example

MarcelCoding: right, no problem. I'll fix it

maxim-lobanov: Looks like PR checks hang for some reason 😄 
I will validate caching as soon as they wake up.

MarcelCoding:
I don't did a force push, that you can see the changes better. I think you can squash this pr at the end. I added also some tests for the toolchain folder for adopt.

maxim-lobanov:
@MarcelCoding validated and it works as expected with cache.
@konradpabjan final look from your side after latest changes :) If you don't have concerns, we can proceed with merging / releasing on Monday

karianna:
Hi all - MMartijn from AdoptOpenJDK here.  With the shift to Eclipse Adoptium we won't be providing OpenJ9 based builds going forward (that'll come directly from IBM).  I don't want to speak directly for IBM here, but if we could pause on that a few days so that they can chime in?  CC @tellison

maxim-lobanov:
Hello @karianna , Yes, I think it is not a problem to wait a few days

bmarwell:
> Hi all - Martijn from AdoptOpenJDK here. With the shift to Eclipse Adoptium we won't be providing OpenJ9 based builds going forward (that'll come directly from IBM). I don't want to speak directly for IBM here, but if we could pause on that a few days so that they can chime in? CC @tellison

I do not think this will happen. In that case IBM would have informed me on my business mail. They know that we are using the adoptopenjdk.net API. Are you sure this hasn’t been forgotten to communicate to customers?

MarcelCoding:
@karianna isn't Openj9 in the eclipse foundation?

karianna:
Hi Marcel,

Yes, Eclipse OpenJ9 is in the Eclipse Foundation, but the Foundation does
not actually build the binaries itself, only AdoptOpenJDK and IBM ever did
as vendors.

Cheers,
Martijn

karianna:
Hi Benjamin,

My apologies I definitely didn't want to cause any concern here!  The
existing Adopt API is definitely *not* going away (if ever) far too many
folks rely on it.

At some stage, the next iteration of Hotspot-based binaries will be served
through the upcoming api.adoptium.net (being worked on as part of the
Adoptium Marketplace concept) and we'll be asking folks to migrate over if
they want to access those newer binaries.

We'll get some clarity on the OpenJ9 story soon, but again to reassure
you, we're *not* switching off the existing API.

----

Right after having written all of this, I think this PR should just go
ahead as it accurately represents what's happening today.  A subsequent PR
can always go in later to model whatever happens in the future.

Cheers,
Martijn

mstoodle:
Here's some recent info about access to Eclipse OpenJ9 based JDKs : https://developer.ibm.com/blogs/ibm-joins-eclipse-adoptium-and-offers-free-certified-jdks-with-eclipse-openj9/

I'm afraid there's not much detail in there to help progress the OpenJ9 part of this PR. As far as I know, though, nothing is changing for the April update release coming very soon at AdoptOpenJDK. We'll be providing more details in advance of the release process migrating fully to Eclipse Adoptium. Stay tuned (but don't hold your breath :) )!

konradpabjan:
Thanks for the info @karianna @mstoodle 

Given that the existing APIs are not being switched over yet (it will probably take some time), I'm inclined to merge these changes in now. Any objections?

The OpenJ9 story is a bit ambiguous at the moment but I don't consider that to be a blocker. If we do need to make changes in the future then I'm sure we'll know well in advance and we'll have plenty of time to react.

lastthyme (approved):

lastthyme (approved):

bmarwell:
Any plans merging this and releasing a v3 @maxim-lobanov?

maxim-lobanov:
Merged PR and released version `v2.1.0` [Link](https://github.com/actions/setup-java/releases/tag/v2.1.0). No need to bump major version since there are no breaking changes.
Going to update `v2` tag to point to `v2.1.0` tomorrow. Will be appreciate for any testing and confirmation that it works as expected.

sehrope:
After updating the action version to `@v2.1.0` I can confirm this worked with both `adopt-openj9` and `adopt-hotspot`.

maxim-lobanov:
Updated `v2` tag

PR #163, opened 2021-04-12
Fix typo on distribution "jdkfile" in Migration Guide

WtfJoke:
**Description:**
There is a small typo.
Its "jdkfile" and not "jdkFile", see https://github.com/actions/setup-java/blob/b53500dabc37db6cc1dcd35059bdcd85df25d140/src/distributions/distribution-factory.ts#L10

To avoid error: No supported distribution was found for input jdkFile

**Related issue:**

**Check list:**


maxim-lobanov (approved):

maxim-lobanov:
Thank you for contribution and fixing it!

PR #171, opened 2021-04-21
Update README.md

vogella:
This updates the Readme examples by using only one space instead of two.

Fixes actions#170

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ steps:
 - uses: actions/checkout@v2
 - uses: actions/setup-java@v2
   with:
-    distribution: 'adopt' # See 'Supported distributions' for available options
-    java-version: '11'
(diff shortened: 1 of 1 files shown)


maxim-lobanov:
Hello @vogella , I don't think that we should replace 2 spaces to the one space.
The suggested syntax for YAML files is to [use 2 spaces for indentation](https://developers.home-assistant.io/docs/documenting/yaml-style-guide/#indentation).
Also it is recommended by [GitHub Actions documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-multiple-events-with-activity-types-or-configuration).

vogella:
THanks for the reply @maxim-lobanov. Closing the issue.

PR #179, opened 2021-05-11
Export main action functions, ease reuse

paulolieuthier:
**Description:**
Export main actions functions to allow wrapper actions that use this action.

**Related issue:**
No filed issue.

**Checklist:**

### DIFF
--- a/dist/cleanup/index.js
+++ b/dist/cleanup/index.js
@@ -2592,6 +2592,7 @@ function run() {
         }
     });
 }
+exports.default = run;
 run();
 
 
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -35713,6 +35713,7 @@ function run() {
         }
     });
 }
+exports.default = run;
 run();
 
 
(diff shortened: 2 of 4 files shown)


konradpabjan:
Hey @paulolieuthier 

These changes look harmless (you also need to run `npm run build` so that all the `.js` files under `dist/*` get updated too. 

I'm having a hard time understanding the use-case for this though 🤔 We have [composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action) that let you build and expand upon existing actions, but this seems like it wouldn't be useful for those scenarios. How could wrapper functions use this? I'm thinking it might be better for anyone to just clone & tweak this action if they want to create some sort of wrapper with extra functionality. Having these changes directly in here aren't that beneficial

paulolieuthier:
> We have composite actions that let you build and expand upon existing actions

AFAIK, this feature only covers script commands. I'm waiting for a new version supporting arbitrary actions: https://github.com/actions/runner/issues/646.

> How could wrapper functions use this?

Basically importing it as a dependency and calling the exported function directly. It's not ideal, especially considering there is no API to set inputs and get outputs.

> you also need to run npm run build so that all the .js files under dist/* get updated too

Totally forgot that.

dmitry-shibanov:
Hello @paulolieuthier. We do not think it makes sense to accept this pull request. Note: The attached issue was resolved. We think it should resolve your issue.

PR #193, opened 2021-07-13
Introduce the dependency caching for Maven and Gradle

KengoTODA:
This PR suggests the dependency caching feature just like the [dependency caching feature of `actions/setup-node`](https://github.blog/changelog/2021-07-02-github-actions-setup-node-now-supports-dependency-caching/).

**Benefits:**
* It makes the workflow YAML file shorter, no need to run `actions/cache` explicitly any more.
* It can run Java installation and cache restoring process in parallel, so elapsed time for initialization will be shorter.

About the performance, we can expect that it'll **shorten about 28% of the time to build**, even when we have no dependency. You can find the code to run the test in my [test project](https://github.com/kengotoda/setup-java-perf-test/), or see [setup-java-cache.csv](https://github.com/actions/setup-java/files/6805263/setup-java-cache.csv) or [Google Docs](https://docs.google.com/spreadsheets/d/1Q7OQJ7UuVNg78FeNhc1DfMAJGJnKJoCHTXQibfMmrlE/edit?usp=sharing) to grab detailed data.

Perf  comparison for Java cache

Note that in this graph I run 10 tests and use 9 of them (treated 1 of them as error) for each dependency size.

**Considerations:**
* This feature does not provide the [`upload-chunk-size` configuration](https://github.com/actions/cache/blob/0638051e9af2c23d10bb70fa9beffcad6cff9ce3/action.yml#L14-L16). I think it won't be problem for most users, and `actions/setup-node` also provides no config like this.
* We may support other package managers like [ivy](https://ant.apache.org/ivy/history/2.5.0/settings/caches.html) and [grape](http://docs.groovy-lang.org/3.0.8/html/documentation/grape.html), but first let's focus on package managers described in the [doc of `actions/cache`](https://github.com/actions/cache/blob/0638051e9af2c23d10bb70fa9beffcad6cff9ce3/examples.md#java---gradle).

**Check list:**

Thanks in advance for your review! 😃


maxim-lobanov:
Hello @KengoTODA , thank you for contribution!
Your changes look really awesome. Please give us some time to review it properly and make sure that everything is okay and we will be happy to accept it.

src/setup-java.ts
```diff
@@ -27,6 +29,7 @@ async function run() {
       throw new Error(`No supported distribution was found for input ${distributionName}`);
     }
 
+    const restoreResult = cache ? restore(cache) : Promise.resolve();
```
dmitry-shibanov, line 32: Could you please make it via if check and move it after configureAuthentication. As I know promise.all keep order for returning of the array but not for execution. I would prefer to configure authentication firstly.

src/cache.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+ * A function that generates a cache key to use.
+ * Format of the generated key will be "${{ platform }}-${{ id }}-${{ fileHash }}"".
+ * If there is no file matched to {@link PackageManager.path}, the generated key ends with a dash (-).
+ * @see {@link https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key|spec of cache key}
+ */
+async function computeCacheKey(packageManager: PackageManager) {
+  return `${process.env['RUNNER_OS']}-${packageManager.id}-${await glob.hashFiles(
```
dmitry-shibanov, line 52: Could you please add to the cache key `setup-java` to look something like this: `setup-java-${process.env['RUNNER_OS']}-${packageManager.id}-${await glob.hashFiles(packageManager.pattern.join('\n'))}`

Nitpick: Could you please split this commands on the lines.

.github/workflows/e2e-cache.yml
```diff
@@ -0,0 +1,88 @@
+name: Validate cache
+on:
+  push:
```
dmitry-shibanov, line 3: Could you please add e2e tests for maven.

.github/workflows/e2e-cache.yml
```diff
@@ -0,0 +1,88 @@
...
+            plugins { id 'java' }
+            repositories { mavenCentral() }
+            dependencies { implementation 'org.codehaus.groovy:groovy:1.8.6' }
+            tasks.register('downloadDependencies') { doLast {
+              def total = configurations.compileClasspath.inject (0) { sum, file ->
+                sum + file.length()
+              }
+              println total
+            }}
+          EOF
```
dmitry-shibanov, line 46: Could please move it to environment variable to reuse it or create a file in __tests__ folder ?

.github/workflows/e2e-cache.yml
```diff
@@ -0,0 +1,88 @@
...
+          if [ -d ~/.gradle/caches ]; then
+            echo "::error::The ~/.gradle/caches directory exists unexpectedly"
+            exit 1
+          fi
```
dmitry-shibanov, line 74: Could you please remove this check.

KengoTODA: Hi, could you tell me your motivation bit more detail? I thought it's necessary to ensure the directory was created by the action. 🤔

src/setup-java.ts
KengoTODA: Let me confirm: is the following change satisfy your request?

```suggestion
    await auth.configureAuthentication();
    await restoreResult;
```

Or do we need to move `restore(cache)` to after `configureAuthentication`? This change probably makes the benefit of this PR less, because we cannot run JDK installation and restoring the dependency cache in parallel.

KengoTODA:
@dmitry-shibanov I've updated the code, please check. 

I also found that the Gradle cache on Windows isn't be saved as expected, see [this build result](https://github.com/KengoTODA/setup-java/runs/3066617130?check_suite_focus=true#step:7:5) for detail. I will investigate.

src/setup-java.ts
KengoTODA: Confirmed that performance isn't good if [we run the restore process after the auth ends](https://github.com/KengoTODA/setup-java/commit/bd074812489119ed6274bc5c17b836e16f3a612a).
Its 95 percentile is [5% slower than `actions/cache + actions/setup-java`](https://docs.google.com/spreadsheets/d/1Q7OQJ7UuVNg78FeNhc1DfMAJGJnKJoCHTXQibfMmrlE/edit#gid=1914328659&range=D37).

So if we need to run the cache after the auth, I think this PR isn't so valuable for performance.

dmitry-shibanov:
> @dmitry-shibanov I've updated the code, please check.
> 
> I also found that the Gradle cache on Windows isn't be saved as expected, see [this build result](https://github.com/KengoTODA/setup-java/runs/3066617130?check_suite_focus=true#step:7:5) for detail. I will investigate.

Hello @KengoTODA. Thank you for your help. It looks like a known issue with gradle on windows-latest. Possibly it will help: https://github.com/actions/cache/issues/454

.github/workflows/e2e-cache.yml
dmitry-shibanov: Hello @KengoTODA. It can be removed because the new job will be started with no `~/.gradle/caches` directory. It will be created after `setup-java` step

morki:
Hi, I see the usage in some test like this:

```yaml
- uses: actions/setup-java@v2
  with:
    distribution: 'adopt'
    java-version: '11'
    cache: 'gradle' # will restore cache of dependencies and wrappers
```

Please will it be possible to use `working-directory` to make it use other Gradle directory than root like this?

```yaml
- uses: actions/setup-java@v2
  working-directory: server
  with:
    distribution: 'adopt'
    java-version: '11'
    cache: 'gradle' # will restore cache of dependencies and wrappers
```

We use monorepo structure, where `server` directory is for Java application and `client` for Node one.
Maybe some other solution can be used to accomplish this.

Thank you very much for such a great feature!

.github/workflows/e2e-cache.yml
KengoTODA: OK I understood, I've removed this process. Thanks!

src/setup-java.ts
KengoTODA: I've rechecked the performance after we fix the cache process on windows, and confirmed that the performance isn't bad; [21% faster than `actions/cache + actions/setup-java`](https://docs.google.com/spreadsheets/d/1Q7OQJ7UuVNg78FeNhc1DfMAJGJnKJoCHTXQibfMmrlE/edit#gid=2134038139&range=D37).

Perf  comparison for Java cache (1)
I'll push 6977c031ae59b16fb87b34271efc3bd5fb24b057 to make the restore process running after the auth.

KengoTODA:
Thanks all,

I've finished the necessary improvements, please check updates here.
Now e2e tests on Windows are stable, we also have tests for Maven, and the restore process runs after the auth part.

src/cache.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+
+/**
+ * Save the dependency cache
+ * @param id ID of the package manager, should be "maven" or "gradle"
+ */
+export async function save(id: string) {
+  const packageManager = findPackageManager(id);
+  const primaryKey = await computeCacheKey(packageManager);
```
dmitry-shibanov, line 93: Hello @KengoTODA. Could you please change `const primaryKey = await computeCacheKey(packageManager);` to `const primaryKey = core.getState(STATE_CACHE_PRIMARY_KEY);`

KengoTODA: OK, I've confirmed that the [`actions/cache` also use the state](https://github.com/actions/cache/blob/5807af2642b6ffc80df306359122fd0ff9b571b8/src/save.ts#L30-L31), and seems nice then I'll change it so. :)
Thank you!

KengoTODA: done by https://github.com/actions/setup-java/pull/193/commits/7fe6c4d63194dfa8e9a1e3c8bdcd5f08b63ad5f9 👍

KengoTODA:
To make the [permission trouble on Windows](#issuecomment-879855343) more intuitive, added a warning like below:

warning-for-gradle-on-windows

dmitry-shibanov (approved):

src/cache.ts
```diff
@@ -0,0 +1,135 @@
...
+ * @param id ID of the package manager, should be "maven" or "gradle"
+ */
+export async function restore(id: string) {
+  const packageManager = findPackageManager(id);
+  const primaryKey = await computeCacheKey(packageManager);
+
+  core.debug(`primary key is ${primaryKey}`);
+  core.saveState(STATE_CACHE_PRIMARY_KEY, primaryKey);
+  if (primaryKey.endsWith('-')) {
+    core.warning(
```
maxim-lobanov, line 68: I wonder if we would like to `throw new Error()` if no files are found matched patterns like we have done in setup-node.

@AlenaSviridenko @MaksimZhukov  what do you think about it?

If we would like to keep it as a warning, I think we will need one more condition in `save` function because current implementation will try to save cache even in case when no files are matched patterns. I guess we don't want it.

src/cleanup-java.ts
```diff
@@ -14,4 +15,39 @@ async function run() {
...
+      .catch(error => {
+        core.warning(error);
+        resolve(void 0);
+      })
+      .then(resolve);
+  });
+}
+
+export async function run() {
+  await Promise.all([removePrivateKeyFromKeychain(), ignoreError(saveCache())]);
```
maxim-lobanov, line 45: I would suggest to run both actions step by step:
```
await removePrivateKeyFromKeychain()
await ignoreError(saveCache())
```

If we run them in parallel, I guess the produced logs will be mixed that can complicate debugging. `removePrivateKeyFromKeychain` is pretty fast step so it is better to run it firstly.

maxim-lobanov:
@KengoTODA, thank you for your contribution. I have left a few minor comments.
There is one more issue that we need to resolve before this pull-request can be merged:
Post-job step does two action: clean up auth and save cache.
[Post-job condition](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#post-if) is `always()` currently. It means that post-job step will be run even job is failed. It is necessary because auth data should be clean up always.
The problem is that we don't want to save cache if job fails because obviously, it will cause invalid caching and save and etc (for example, when gradle install fails).
I didn't find the way to determine if the whole job is successful or not in post job step. Will try to ask our engineers if there is a way to overcome it.

Please let me know if you have any suggestions / concerns here.

src/cache.ts
KengoTODA: Thanks for your review! I just kept the existing behaviour (`@actions/cache` with `hashFiles(...)`), but it would be a great improvement making cache more intuitive. :+1:

I'll wait for the reply from other experts, and change the code once we made a consensus.

src/cleanup-java.ts
KengoTODA: Fixed, thanks!

dmitry-shibanov:
Hello @KengoTODA. Sorry for the late reply. Could you please take a look on [pull request](https://github.com/KengoTODA/setup-java/pull/3) we've created in your repository ? We've added logic to save cache only for successful builds. 
Besides, could you please pull latest main to resolve the conflict ?

KengoTODA:
@dmitry-shibanov Thanks for your PR! I've merged it and resolved a conflict between the upstream repo. Please check :)

src/cache.ts
maxim-lobanov: + @konradpabjan for any thoughts

maxim-lobanov:
@KengoTODA, sorry for delay and thank you for your patience. Issue with post-job required to apply "hacks" :) 
I think we are on the final line with this PR and it will be merged soon. We will double validate from our side and engage a few more experts to review.

maxim-lobanov (approved):

vsafonkin (approved):

src/util.ts
```diff
@@ -69,3 +69,9 @@ export function getToolcachePath(toolName: string, version: string, architecture
 
   return null;
 }
+
+export function isJobStatusSuccess() {
+  const jobStatus = core.getInput('job-status');
```
konradpabjan, line 74: Could you also add `job-status` as a constant to `src/constants.ts` (maybe `INPUT_JOB_STATUS`). Minor but lets follow the pattern we have for all inputs.

konradpabjan (commented):
Awesome PR and thank you for the contribution! 🎉 

Left a few minor comments.

The changes to the `index.js` files under `/dist` are also huge (pretty much impossible to manually verify). I've created a separate PR to add a automated check to make it easier to verify the files are correct. Want to do a bit more verification once https://github.com/actions/setup-java/pull/206 goes in.

src/cache.ts
KengoTODA: I made a proposal in my repo: https://github.com/KengoTODA/setup-java/commit/530d597222235536edde1012c73bc9034f8a1f97

If it looks good to you, I'll merge this commit to this PR. :)

maxim-lobanov: @KengoTODA , Proposal looks good to me. Please merge PR.

Also, I see that two checks fail right now:
- license - please ignore this check for now. I will fix it in the separate PR.
- build fails on verify_no_unstaged_changes - may be you haven't rebuild it last time? Let's try to rebuild one more time after merging your proposal via `npm run build && npm run release`

KengoTODA:
I've updated licensed files with [licensed v3.1.0](https://github.com/github/licensed/releases/tag/3.1.0). It still has three status check failures:

>* setup-java.npm.sax
>  filename: /Users/kengo/GitHub/setup-java/.licenses/npm/sax.dep.yml
>    - license needs review: other
>
>* setup-java.npm.tslib-1.14.1
>  filename: /Users/kengo/GitHub/setup-java/.licenses/npm/tslib-1.14.1.dep.yml
>    - license needs review: 0bsd
>
>* setup-java.npm.tslib-2.3.0
>  filename: /Users/kengo/GitHub/setup-java/.licenses/npm/tslib-2.3.0.dep.yml
>    - license needs review: 0bsd

The [license of sax v1.2.4 is ISC](https://github.com/isaacs/sax-js/blob/v1.2.4/package.json#L7). And all of them are already used in the `setup-node` action, so I hope there is no problem using them in this project:

* https://github.com/actions/setup-node/blob/v2.4.0/.licenses/npm/sax.dep.yml
* https://github.com/actions/setup-node/blob/v2.4.0/.licenses/npm/tslib-1.14.1.dep.yml
* https://github.com/actions/setup-node/blob/v2.4.0/.licenses/npm/tslib-2.3.0.dep.yml

maxim-lobanov:
@KengoTODA , 
As for the licenses:
- Change `sax.dep.yml` to set `license: isc` manually.
- Add `0bsd` to https://github.com/actions/setup-java/blob/main/.licensed.yml#L4 as it is done for setup-node: https://github.com/actions/setup-node/blob/v2.4.0/.licensed.yml

Could you please also merge your proposal https://github.com/KengoTODA/setup-java/pull/4 and `npm run build && npm run release` after that?

KengoTODA:
Here is (probably) the final performance comparison chart. As expected, the performance is better than existing way:

Perf  comparison for Java cache (67b74a15) (1)

Source (workflow runs): [large](https://github.com/KengoTODA/setup-java-perf-test/actions/runs/1146837665), [medium](https://github.com/KengoTODA/setup-java-perf-test/actions/runs/1146837666), and [small](https://github.com/KengoTODA/setup-java-perf-test/actions/runs/1146837668)

konradpabjan (approved):
🚀 🥳 LGTM!

maxim-lobanov:
@KengoTODA thank you one more time for great contribution, very appreciate!
I have merged PR and changes are already in `main` branch.
I will cut the new action release on Monday (we are trying to avoid releases on Friday)

KengoTODA:
I'm very glad to see a [changelog entry announcing this change](https://github.blog/changelog/2021-08-30-github-actions-setup-java-now-supports-dependency-caching/). Thanks again everyone! 🚀

PR #198, opened 2021-07-20
Add problem matcher for Checkstyle

### LINKED ISSUE #197: Problem matcher for Checkstyle (and possibly others)
**Description:**
Problem matchers for tools such as [Checkstyle](https://checkstyle.sourceforge.io/) (and possibly others).

**Justification:**
Allows annotations on the GitHub interface. It is nicer if these are included in the `setup-java` repo rather than another action. This is available for some actions, such as [`setup-node`](https://github.com/actions/setup-node/blob/main/.github/eslint-stylish.json).

Yes, if you would accept one.

kiancross:
**Description:**

This PR adds a problem matcher for Checkstyle to allow automatic GitHub annotations.

When run through Gradle, an example line is:
```
[ant:checkstyle] [ERROR] /root/Bound.java:7:2: 'import' has incorrect indentation level 1, expected level should be 0. [Indentation]
```

Alternatively, when run standalone, an example output is:
```
[WARN] /root/test.java:2:1: The name of the outer type and the file do not match. [OuterTypeFilename]
```

The regular expression matches both cases.

In the latter example, `WARN`, rather than `WARNING` is output. This is resolved by using a default severity of `warning`, which should be overridden in the case that the severity is `ERROR`.

**Related issue:**
Closes #197.

**Check list:**

### DIFF
--- a/.github/checkstyle.json
+++ b/.github/checkstyle.json
@@ -1,0 +1,19 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "checkstyle",
+            "severity": "warning",
+            "pattern": [
+                {
+                    "regexp": "^.*\\[(ERROR|WARN)\\]\\s+(.*):(\\d+):(\\d+):\\s+(.*)\\s+\\[(.*)\\]$",
+                    "severity": 1,
+                    "file": 2,
+                    "line": 3,
+                    "column": 4,
+                    "message": 5,
+                    "code": 6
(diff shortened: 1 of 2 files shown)


kiancross:
I have created a separate action with this instead. Feel free to copy into this repo if you prefer.

https://github.com/kiancross/java-annotations-checkstyle

PR #206, opened 2021-08-19
Add check-dist.yml

konradpabjan:
This adds a workflow to check the contents of the checked-in dist/index.js against the expected version.

See actions/upload-artifact#227 for background and rationale.

Similar PR: https://github.com/actions/download-artifact/pull/108

### DIFF
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -1,0 +1,52 @@
+# `dist/index.js` is a special file in Actions.
+# When you reference an action with `uses:` in a workflow,
+# `index.js` is the code that will run.
+# For our project, we generate this file through a build process
+# from other source files.
+# We need to make sure the checked-in `index.js` actually matches what we expect it to be.
+name: Check dist/
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


KengoTODA:
note that we have similar comparison in the `build.yml`:
https://github.com/actions/setup-java/blob/4b1b3d4a82f726de9ba1945f8c6967b67caf5dcd/.github/workflows/build.yml#L31-L33

It runs [verify-no-unstaged-changes.sh](https://github.com/actions/setup-java/blob/4b1b3d4a82f726de9ba1945f8c6967b67caf5dcd/__tests__/verify-no-unstaged-changes.sh) in the `__tests__` dir.

maxim-lobanov:
@konradpabjan , agree with @KengoTODA . In `build.yml` we do pretty similar check: https://github.com/actions/setup-java/blob/main/__tests__/verify-no-unstaged-changes.sh
I think it should be enough.

konradpabjan:
🤦‍♂️ Disregard. Thanks everyone!

PR #212, opened 2021-08-25
Create check-dist.yml

brcrista:
This adds a workflow to check the contents of the checked-in `dist/index.js` against the expected version.

See https://github.com/actions/upload-artifact/pull/227 for background and rationale.


brcrista:
@konradpabjan @maxim-lobanov @KengoTODA I think there's value in using the same mechanism for all 1st-party actions.  For `cache` we moved its existing step in the build job to its own check: https://github.com/actions/cache/pull/618

Though I think the value is just in consistency, I don't think same job vs. different job is a huge deal.

dmitry-shibanov:
Hello @brcrista. Thank you for your pull request. Could you please remove this [step](https://github.com/actions/setup-java/blob/3bc31aaf88e8fc94dc1e632d48af61be5ca8721c/.github/workflows/build.yml#L31-L33) and the [verify-no-unstaged-changes.sh](https://github.com/actions/setup-java/blob/main/__tests__/verify-no-unstaged-changes.sh) file.

PR #213, opened 2021-08-26
feat(cache): change key to match actions/cache documentation

mmorel-35:
**Description:**
This PR change the cache key construction to fit [those examples](https://github.com/actions/cache/blob/main/examples.md#java---gradle). It also add the use of restoreKeys and a more verbose log when retrieving fails.

It also add the pre-checkin npm task as it exists on the other setup-X

I don't see how to fix this but the "Verify no unstaged changes" is failing. I'm working on Windows but usually it doesn't matter.
Any idea ?

**Related issue:**
N/A

**Check list:**
- [X] Mark if tests were added or updated to cover the changes.


KengoTODA:
> I don't see how to fix this but the "Verify no unstaged changes" is failing. I'm working on Windows but usually it doesn't matter.

It is caused by the difference between LF (Ubuntu) and CRLF (Windows). I also faced this problem before.
I solved it by replacing the `dist` with files built on linux machine, we can upload the `dist` file to workflow by `actions/upload-artifact`:

https://github.com/KengoTODA/setup-java/blob/aae89bdd784d68e53c1bda327075e7af68d50ac5/.github/workflows/build.yml#L31-L34

KengoTODA:
Note about the context of current key naming:

1. First, it followed the actions/cache documentation: `${process.env['RUNNER_OS']}-${packageManager.id}-${hash}`
2. Second, [get reviewed and suggested to add a prefix](https://github.com/actions/setup-java/pull/193#discussion_r669521434), so
3. [Added a prefix](https://github.com/KengoTODA/setup-java/commit/f537562e33870bacc822d7489bbc0ff977f91ad0#diff-9d9b93c9a37cd43e82ab7d4127eb0c21181f5eb00c65673f7a605afd94097248) and now it's `${CACHE_KEY_PREFIX}-${process.env['RUNNER_OS']}-${packageManager.id}-${hash}`

mmorel-35:
Hi @KengoTODA, I found another, to correct the file, I'm creating a workflow where I execute the build and then have a task to commit the changed file directly on my repository.

Concerning the notion of prefix I'm not sure to really understand the reason behind that. Does it means that the provided examples in actions/cache are wrong ?  Their is no explanation behind that change in the comment. 
What I'm doing here is just moving this prefix after the operating system which quite more common in the community as most of the people take example on that same page I referenced earlier.

KengoTODA:
> Does it means that the provided examples in actions/cache are wrong ?

It's not wrong. And there is no strict official rule to design the cache key... what we need is just grab [how actions/cache matches a cache key](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) and design optimized key and restore keys. So I just shared the context of current change... repository owner will judge how the key should be designed.

fabiosanto (approved):

PR #215, opened 2021-08-31
Describe cache key format in readme

beatngu13:
**Description:**
Several users, including myself, wondered what cache key format is being used under the hood (see e.g. [here](https://twitter.com/lorenzo_bettini/status/1432580351491194880) or [here](https://twitter.com/brenuart/status/1432623680333488131)). This PR adds a corresponding description – based on [`computeCacheKey`](https://github.com/actions/setup-java/blob/main/src/cache.ts#L52) – to the `README.md`.

**Related issue:**
n/a

**Check list:**
- [ ] Mark if documentation changes are required.
- [ ] Mark if tests were added or updated to cover the changes.

PR #219, opened 2021-09-06
Discussion: Is it better to remove OS from the cache key?

KengoTODA:
**Description:**

This PR is for discussion about the `RUNNER_OS` in the cache key.

I found an opinion at Twitter: [it's needless because Java's package are portable](https://twitter.com/rmannibucau/status/1432579398415392769), and I agree with it.
In my understanding, the Maven repository provides no feature to serve different files to different OSs.

I prepared necessary changes to hold discussion with you. How do you think about this change?
It may make the build efficient in some case, especially when we run builds on multiple OSs in serial.

**Related issue:**

* #213 
* #215

**Check list:**
- [x] Mark if documentation changes are required; This PR requires updates in the change from #215 

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -45,13 +45,13 @@ export function findPackageManager(id: string): PackageManager {
 
 /**
  * A function that generates a cache key to use.
- * Format of the generated key will be "${{ platform }}-${{ id }}-${{ fileHash }}"".
+ * Format of the generated key will be "setup-java-${{ id }}-${{ fileHash }}"".
  * If there is no file matched to {@link PackageManager.path}, the generated key ends with a dash (-).
  * @see {@link https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key|spec of cache key}
  */
 export async function computeCacheKey(packageManager: PackageManager) {
   const hash = await glob.hashFiles(packageManager.pattern.join('\n'));
-  return `${CACHE_KEY_PREFIX}-${process.env['RUNNER_OS']}-${packageManager.id}-${hash}`;
+  return `${CACHE_KEY_PREFIX}-${packageManager.id}-${hash}`;
 }
 
 /**
@@ -73,7 +73,7 @@ export async function restore(id: string) {
   }
 
   const matchedKey = await cache.restoreCache(packageManager.path, primaryKey, [
-    `${CACHE_KEY_PREFIX}-${process.env['RUNNER_OS']}-${id}`
+    `${CACHE_KEY_PREFIX}-${id}`
   ]);
   if (matchedKey) {
     core.saveState(CACHE_MATCHED_KEY, matchedKey);
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -35,7 +35,7 @@ const supportedPackageManager: PackageManager[] = [
   }
 ];
 
-function findPackageManager(id: string): PackageManager {
+export function findPackageManager(id: string): PackageManager {
   const packageManager = supportedPackageManager.find(packageManager => packageManager.id === id);
   if (packageManager === undefined) {
     throw new Error(`unknown package manager specified: ${id}`);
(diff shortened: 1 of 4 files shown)


src/cache.ts
```diff
@@ -45,13 +45,13 @@ function findPackageManager(id: string): PackageManager {
...
- * Format of the generated key will be "${{ platform }}-${{ id }}-${{ fileHash }}"".
+ * Format of the generated key will be "setup-java-${{ id }}-${{ fileHash }}"".
  * If there is no file matched to {@link PackageManager.path}, the generated key ends with a dash (-).
  * @see {@link https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key|spec of cache key}
  */
-async function computeCacheKey(packageManager: PackageManager) {
+export async function computeCacheKey(packageManager: PackageManager) {
   const hash = await glob.hashFiles(packageManager.pattern.join('\n'));
-  return `${CACHE_KEY_PREFIX}-${process.env['RUNNER_OS']}-${packageManager.id}-${hash}`;
+  return `${CACHE_KEY_PREFIX}-${packageManager.id}-${hash}`;
```
rmannibucau, line 54: Guess it should be configurable.
As mentionned on twitter, it is rare but sometimes needed (native like jansi/netty/...).
Default shouldnt use it IMHO but should be possible to have it.
Alternatively, cache creation can check binary (.so) in jar presence maybe?

KengoTODA: > it is rare but sometimes needed (native like jansi/netty/...).

ah I got it, Maven repo always serves same files but Maven project may use build profile to add dependencies only to specific OS. Gradle project can do similar build too.

In my current opinion, it's hard to provide a smart and high-speed feature detecting such dependency on OS.  😢
Then next consideration is how to make it configurable; but it may make the workflow file complex and nonintuitive. Here I imagine new settings like `cache: os-dependent-maven` and `cache-key-strategy: os-dependent`.

So I suggest making cache key OS independent for major use cases, and ask rare case users keep using `actions/cache` to make the deps cache OS dependent. Any thought?

Thanks again for making the product better with me 🙌

rmannibucau: as an user, I would expect the composite actions to expose the underlying composed actions and just - optionally - provide nice defaults if possible.
So overall it would be something like `cache: <actions/cache configuration to merge with defaults of this action (lodash like merge)>`.
If done, the key can stay OS dependent by default while the doc explicitly state that and shows how to drop the os specific part IMHO.
It will also enable to exclude or not part of the files (a common example is to exclude snapshots) quite easily.
So at the end, the feature can also be, to not have the cache by default until cache subaction is enabled by setting at least a key.
A warning can be logged if not done saying the build can be slow.
Something along those lines wouldn't shock me to be honest.

KengoTODA: > So overall it would be something like cache: <actions/cache configuration to merge with defaults of this action (lodash like merge)>.

Like below?

```yaml
- uses: actions/setup-java@v2
   with:
     java-version: 11
     distribution: temurin
     cache:
       path: ~/.m2/repository
       key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
       restore-keys: |
         ${{ runner.os }}-maven-
```

I don't know that the Actions support nested (object type) input, though.
It's surely intuitive but brings limited benefit to shorten the workflow file, I think.

rmannibucau: yes
The benefit is to override the defaults properly (typically path including the repo but NOT *-SNAPSHOT for ex and key using pom.xml properly).
Overall point is that if the benefit is low, the benefit to include the cache should be the same - but I strongly think it is good to encourage it.
Maybe just checking the yaml has cache action defined it a good outcome instead of trying to merge them.
Will also be saner for all hybrid builds (we have more and more npm+mvn builds these days).

maxim-lobanov: We don't have plans to add any customization of caching in scope of actions/setup-java. If you need customization, it is recomended to use [actions/cache](https://github.com/actions/cache) directly.
We had discussion about it when implemented caching for Node.js: https://github.com/actions/setup-node/blob/main/docs/adrs/0000-caching-dependencies.md#goals--anti-goals

> We don't pursue the goal to provide wide customization of caching in scope of actions/setup-node action. The purpose of this integration is covering ~90% of basic use-cases. If user needs flexible customization, we should advice them to use actions/cache directly.

As for the removing platform from key, we will investigate it deeper

KengoTODA:
This PR is stale for months, and there is no more discussions on this topic so I will close this PR. Thanks all for your help! :)

PR #252, opened 2021-11-18
Add microsoft distribution of the JDK.

brendandburns:
**Description:**
Initial add of the Microsoft OpenJDK distribution

**Related issue:**
N/A

**Check list:**
- [X] Mark if documentation changes are required.
- [X] Mark if tests were added or updated to cover the changes.

PR #259, opened 2021-12-08
Fix the Licensed workflow

brcrista:
**Description:**
The Licensed workflow is failing in https://github.com/actions/setup-java/pull/252.  I don't see any changes there that would cause it to fail.  This PR is to make sure it still passes on `main` and to fix any issues.

**Related issue:**

**Check list:**

**Testing**
Red run before fix: https://github.com/actions/setup-java/actions/runs/1555484816
Green run after fix: https://github.com/actions/setup-java/actions/runs/1555526992

### DIFF
--- a/.github/workflows/licensed.yml
+++ b/.github/workflows/licensed.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


brcrista:
Confirmed that this fixes the workflow: https://github.com/actions/setup-java/actions/runs/1555526992

konradpabjan (approved):

PR #263, opened 2021-12-13
Add an error condition for an unstable Microsoft OpenJDK build

dmitry-shibanov:
**Description:**
In scope of this pull request we add condition to throw an error for an unstable `Microsoft OpenJDK` build. Some distributions support early access versions with syntax `11-ea`. For now `Microsoft OpenJDK` does not provide `ea` versions. The action will not throw an error if user tries to install `ea` version for `Microsoft OpenJDK`

**Related issue:**

**Check list:**

### DIFF
--- a/src/distributions/microsoft/installer.ts
+++ b/src/distributions/microsoft/installer.ts
@@ -40,6 +40,11 @@ export class MicrosoftDistributions extends JavaBase {
     if (this.architecture !== 'x64' && this.architecture !== 'aarch64') {
       throw new Error(`Unsupported architecture: ${this.architecture}`);
     }
+
+    if (!this.stable) {
+      throw new Error('Unstable versions are not supported');
+    }
+
     const availableVersionsRaw = await this.getAvailableVersions();
 
     const opts = this.getPlatformOption();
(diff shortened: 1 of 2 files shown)


MaksimZhukov (approved):

dmitry-shibanov:
Hello everyone. Could you please take a look at these changes.
cc: @brcrista @brendandburns

src/distributions/microsoft/installer.ts
```diff
@@ -40,6 +40,11 @@ export class MicrosoftDistributions extends JavaBase {
     if (this.architecture !== 'x64' && this.architecture !== 'aarch64') {
       throw new Error(`Unsupported architecture: ${this.architecture}`);
     }
+
+    if (!this.stable) {
+      throw new Error('Unstable versions are not supported');
```
brcrista, line 45: What's the value of adding a special error message here?  We only have two hardcoded versions currently.  If someone requests something that's not one of those versions, we'll raise an error and show the available versions (lines 62-70 of this file).

brcrista:
```suggestion
      throw new Error('Early access versions are not supported');
```

While it's true that it's not a stable version, "unstable" has a stronger denotation.

brcrista (approved):

PR #274, opened 2022-01-13
Include *.kt files in buildSrc in cache key

schuenadel:
**Description:**
When using Gradle with the Kotlin DSL some people use Kotlin files like `Versions.kt` or `Dependencies.kt` in the `buildSrc` folder to define their dependency versions (see [example here](https://github.com/google/iosched/blob/6f96c1528c00d3d89bedb33e5a0a10d67629e8bd/buildSrc/src/main/java/Versions.kt) and general info about [buildSrc here](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources)). 
At the moment a change in these files would not be covered by the file hash in the cache key calculation, which means an updated dependency would be missed and so not been cached.

This PR adds the files `buildSrc/**/*.kt` to the cache key calculation (in addition to the existing `**/*.gradle*`, `**/gradle-wrapper.properties`)

**Related issue:**
None

**Check list:**

Link to test run containing these changes: https://github.com/schuenadel/setup-java/actions/runs/1692869040


vlsi:
Please, no.
Kotlin in `buildSrc` is used for much more than "versions", and using `Versions.kt` in `buildSrc` is an anti-pattern.

If you include `*.kt`, then you basically break cacheability for everybody 🤷

schuenadel:
Hi @vlsi,

> Kotlin in buildSrc is used for much more than "versions"

Yes, I was coming from my "personal Android bubble" where I see it mainly being used for dependencies and if for something else, then rather rarely touched and if so it maybe even rectify a cache miss. But I get your point that this may be different for other, maybe more infrastructure related, projects where the refresh would not be desired.

> and using Versions.kt in buildSrc is an anti-pattern.

I wasn't aware of that and also don't know why it is (I'd say invalidating the build cache on changes doesn't really count for clean builds on CI, if that's the reason...). But anyway many people use it and I think it would be nice to enable them to also benefit from a working cache mechanism.

I preferred to not hardcode the file names as that forces you to use these specific names in order to make the cache work, but maybe that's the better compromise to not have unwanted cache invalidations (as you pointed out). 
So my suggestion would be to use `buildSrc/**/Versions.kt` and `buildSrc/**/Dependencies.kt` instead of `buildSrc/**/*.kt`

schuenadel:
Hi @vlsi (and others),
I added my suggestion

> So my suggestion would be to use buildSrc/**/Versions.kt and buildSrc/**/Dependencies.kt instead of buildSrc/**/*.kt

to restrict the scope of the cache key to the most used(?) files for dependency management. WDYT?

dsame (approved):

marko-zivic-93 (approved):

schuenadel:
Could this be merged, or is something missing? I am lacking rights for merging.

Issue #276, opened 2022-01-13
Maven Toolchains Support


Okeanos:
**Description:**
This is related to #44, however, while that issue in particular asks for multiple JDKs to be installed with a single `setup-java` invocation, I would like the general `setup-java` invocation to provide a [Maven Toolchains](https://maven.apache.org/guides/mini/guide-using-toolchains.html) declaration for the selected JDK(s).

This requires putting a `toolchains.xml` file with the proper definitions into the `$HOME/.m2` location of the current user, i.e. the one running Maven.

Implementation of #44 would require changes to this, i.e. all additional JDKs would have to be correctly exposed in the toolchains declaration as well.

**Justification:**
This would allow using `setup-java` with less overhead in projects that have Maven toolchain requirements configured and also allow Gradle to pick up JDK locations automatically starting with [7.4](https://github.com/gradle/gradle/pull/19048). This was something discussed in #44 as well and is to be implemented in gradle/gradle#14903 with a different solution.

**Are you willing to submit a PR?**
There's a pull request: #282

dmitry-shibanov:
Hello @Okeanos. Thank you for your report. We'll investigate this feature request.

devminded:
Related to #245

paoliniluis:
tools.deps would be a great one to support to, specially for Clojure based projects

brcrista:
To set context, I'm not very familiar with Maven.  That said, would it be possible today to handle Maven configuration in a separate step or action?  If not, is there some sort of output we can add to `setup-java` to make that possible?

Okeanos:
> To set context, I'm not very familiar with Maven. That said, would it be possible today to handle Maven configuration in a separate step or action? If not, is there some sort of output we can add to `setup-java` to make that possible?

That depends a little – the [`toolchains.xml`](https://maven.apache.org/guides/mini/guide-using-toolchains.html) declaration works independently from a particular Maven installation and can also be used by [Gradle](https://docs.gradle.org/current/userguide/toolchains.html#sec:auto_detection) to streamline detection of available JDKs.

To generate a useful and valid toolchains declaration for a particular JDK the following information is necessary:

- JDK version (major version only, e.g. `1.6`, `8`, `16`, `17`)
- JDK location (`/path/to/jdk`)

Optionally, some additional useful information needs to be made available – in particular important for `setup-java` because it offers choices with side-effects:

- JDK vendor
- some sort of stable unique ID to identify this particular toolchain

[...]

Issue #44, opened 2020-01-21
Support multiple JDKs during build


vy:
Certain projects might require multiple JDKs during build (e.g., [Apache Log4j](/apache/logging-log4j2) requires Java 7 and Java 9+ by means of `toolchains.xml`) and in its current form, that is not possible via `setup-java`.

giltene:
+1

It appears to me that adding an optional java-home variable that would determine where to place the java package (which can currently be a jre, jdk, or jdk+fx), overriding the default location, would suffice, right? One would the use setup-java multiple times, once per version.

jvalkeal:
I'm just wondering what happens if you define two different `setup-java` steps with different jvm's. Looking sources it set's this `extendedJavaHome` env variable:

https://github.com/actions/setup-java/blob/b52cd69bd2c048f71381fd0ee11739418a05a362/src/installer.ts#L76-L79

mprins:
I managed to do this by calling out to the `setup-java` action for each JDK I have confgured in my toolchain (1.8 and 11) plus once for the JDK I want to run maven with (so 3 times  in total).

see https://github.com/B3Partners/kadaster-gds2/pull/47

giltene:
> I'm just wondering what happens if you define two different `setup-java` steps with different jvm's. Looking sources it set's this `extendedJavaHome` env variable:

I think that knowing where the JDK was installed is not enough. Some tools expect that java versions to appear at specific file paths. E.g. https://github.com/apache/logging-log4j2/blob/64a5c73636cef24c5cbb70143eef99536944d188/.travis-toolchains.xml#L28 . You could probably do something by symlinking the requested path to where extendedJavaHome points, but that would be messy for the user, and fairly easy for us to do in installer.ts if indicated by setting an optional java-home variable

mprins:
> Some tools expect that java versions to appear at specific file paths.

 see:  
https://github.com/B3Partners/kadaster-gds2/runs/415167771?check_suite_focus=true#step:7:14
```
 Inspect environment0s
JAVA_HOME_8_x64=/opt/hostedtoolcache/jdk/8.0.242/x64
Run set | grep JAVA
  set | grep JAVA
  shell: /bin/bash -e {0}
  env:
    JAVA_HOME: /opt/hostedtoolcache/jdk/8.0.242/x64
    JAVA_HOME_11.0.6_x64: /opt/hostedtoolcache/jdk/11.0.6/x64
[...]

PR #282, opened 2022-01-17
Add Maven Toolchains Declaration (#276)

Okeanos:
**Description:**
* Add Maven Toolchains Declaration after JDK is installed
* Extract common/shared Maven constants

**Related issue:**
#276 

**Check list:**

---

I also used my fork + branch on a private repository's pipeline to verify a valid toolchains declaration is created for the specified JDK and used by a Maven based build job.

Technically, using `${env.JAVA_HOME}` inside the `toolchains.xml` file is possible, however, this will get interesting with #44 very fast. Not to mention that e.g. version and vendor are not exposed as variables anyway.

### DIFF
--- a/__tests__/toolchains.test.ts
+++ b/__tests__/toolchains.test.ts
@@ -1,0 +1,292 @@
+import * as fs from 'fs';
+import os from 'os';
+import * as path from 'path';
+import * as core from '@actions/core';
+import * as io from '@actions/io';
+import * as toolchains from '../src/toolchains';
+import { M2_DIR, MVN_TOOLCHAINS_FILE } from '../src/constants';
+
+const m2Dir = path.join(__dirname, M2_DIR);
+const toolchainsFile = path.join(m2Dir, MVN_TOOLCHAINS_FILE);
+
+describe('toolchains tests', () => {
+  let spyOSHomedir: jest.SpyInstance;
+  let spyInfo: jest.SpyInstance;
+
+  beforeEach(async () => {
+    await io.rmRF(m2Dir);
+    spyOSHomedir = jest.spyOn(os, 'homedir');
+    spyOSHomedir.mockReturnValue(__dirname);
+    spyInfo = jest.spyOn(core, 'info');
+    spyInfo.mockImplementation(() => null);
+  }, 300000);
+
+  afterAll(async () => {
+    try {
+      await io.rmRF(m2Dir);
+    } catch {
+      console.log('Failed to remove test directories');
+    }
+    jest.resetAllMocks();
+    jest.clearAllMocks();
+    jest.restoreAllMocks();
+  }, 100000);
+
+  it('creates toolchains.xml in alternate locations', async () => {
+    const jdkInfo = {
+      version: '17',
+      vendor: 'Eclipse Temurin',
+      id: 'temurin_17',
+      jdkHome: '/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.1-12/x64'
+    };
+
+    const altHome = path.join(__dirname, 'runner', 'toolchains');
+    const altToolchainsFile = path.join(altHome, MVN_TOOLCHAINS_FILE);
+    await io.rmRF(altHome); // ensure it doesn't already exist
+
+    await toolchains.createToolchainsSettings({
+      jdkInfo,
+      settingsDirectory: altHome,
+      overwriteSettings: true
+    });
+
+    expect(fs.existsSync(m2Dir)).toBe(false);
+    expect(fs.existsSync(toolchainsFile)).toBe(false);
+
+    expect(fs.existsSync(altHome)).toBe(true);
+    expect(fs.existsSync(altToolchainsFile)).toBe(true);
+    expect(fs.readFileSync(altToolchainsFile, 'utf-8')).toEqual(
+      toolchains.generateToolchainDefinition(
+        '',
+        jdkInfo.version,
+        jdkInfo.vendor,
+        jdkInfo.id,
+        jdkInfo.jdkHome
+      )
+    );
+
+    await io.rmRF(altHome);
+  }, 100000);
+
+  it('creates toolchains.xml with minimal configuration', async () => {
+    const jdkInfo = {
+      version: '17',
+      vendor: 'Eclipse Temurin',
+      id: 'temurin_17',
+      jdkHome: '/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.1-12/x64'
+    };
+
+    const result = `<?xml version="1.0"?>
+<toolchains xmlns="https://maven.apache.org/TOOLCHAINS/1.1.0"
+  xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="https://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd">
+  <toolchain>
+    <type>jdk</type>
+    <provides>
+      <version>17</version>
+      <vendor>Eclipse Temurin</vendor>
+      <id>temurin_17</id>
+    </provides>
+    <configuration>
+      <jdkHome>/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.1-12/x64</jdkHome>
+    </configuration>
+  </toolchain>
+</toolchains>`;
+
+    await toolchains.createToolchainsSettings({
+      jdkInfo,
+      settingsDirectory: m2Dir,
+      overwriteSettings: true
+    });
+
+    expect(fs.existsSync(m2Dir)).toBe(true);
+    expect(fs.existsSync(toolchainsFile)).toBe(true);
+    expect(fs.readFileSync(toolchainsFile, 'utf-8')).toEqual(
+      toolchains.generateToolchainDefinition(
+        '',
+        jdkInfo.version,
+        jdkInfo.vendor,
+        jdkInfo.id,
+        jdkInfo.jdkHome
+      )
+    );
+    expect(
+      toolchains.generateToolchainDefinition(
+        '',
+        jdkInfo.version,
+        jdkInfo.vendor,
+        jdkInfo.id,
+        jdkInfo.jdkHome
+      )
+    ).toEqual(result);
+  }, 100000);
+
+  it('reuses existing toolchains.xml files', async () => {
+    const jdkInfo = {
+      version: '17',
+      vendor: 'Eclipse Temurin',
+      id: 'temurin_17',
+      jdkHome: '/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.1-12/x64'
+    };
+
+    const originalFile = `<toolchains>
+        <toolchain>
+            <type>jdk</type>
+        <provides>
+        <version>1.6</version>
+        <vendor>Sun</vendor>
+        <id>sun_1.6</id>
+        </provides>
+        <configuration>
+        <jdkHome>/opt/jdk/sun/1.6</jdkHome>
+        </configuration>
+        </toolchain>
+      </toolchains>`;
+    const result = `<?xml version="1.0"?>
+<toolchains>
+  <toolchain>
+    <type>jdk</type>
+    <provides>
+      <version>1.6</version>
+      <vendor>Sun</vendor>
+      <id>sun_1.6</id>
+    </provides>
+    <configuration>
+      <jdkHome>/opt/jdk/sun/1.6</jdkHome>
+    </configuration>
+  </toolchain>
+  <toolchain>
+    <type>jdk</type>
+    <provides>
+      <version>17</version>
+      <vendor>Eclipse Temurin</vendor>
+      <id>temurin_17</id>
+    </provides>
+    <configuration>
+      <jdkHome>/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.1-12/x64</jdkHome>
+    </configuration>
+  </toolchain>
+</toolchains>`;
+
+    fs.mkdirSync(m2Dir, { recursive: true });
+    fs.writeFileSync(toolchainsFile, originalFile);
+    expect(fs.existsSync(m2Dir)).toBe(true);
+    expect(fs.existsSync(toolchainsFile)).toBe(true);
+
+    await toolchains.createToolchainsSettings({
+      jdkInfo,
+      settingsDirectory: m2Dir,
+      overwriteSettings: true
+    });
+
+    expect(fs.existsSync(m2Dir)).toBe(true);
+    expect(fs.existsSync(toolchainsFile)).toBe(true);
+    expect(fs.readFileSync(toolchainsFile, 'utf-8')).toEqual(
+      toolchains.generateToolchainDefinition(
+        originalFile,
+        jdkInfo.version,
+        jdkInfo.vendor,
+        jdkInfo.id,
+        jdkInfo.jdkHome
+      )
+    );
+    expect(
+      toolchains.generateToolchainDefinition(
+        originalFile,
+        jdkInfo.version,
+        jdkInfo.vendor,
+        jdkInfo.id,
+        jdkInfo.jdkHome
+      )
+    ).toEqual(result);
+  }, 100000);
+
+  it('does not overwrite existing toolchains.xml files', async () => {
+    const jdkInfo = {
+      version: '17',
+      vendor: 'Eclipse Temurin',
+      id: 'temurin_17',
+      jdkHome: '/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.1-12/x64'
+    };
+
+    const originalFile = `<toolchains>
+        <toolchain>
+            <type>jdk</type>
+        <provides>
+        <version>1.6</version>
+        <vendor>Sun</vendor>
+        <id>sun_1.6</id>
+        </provides>
+        <configuration>
+        <jdkHome>/opt/jdk/sun/1.6</jdkHome>
+        </configuration>
+        </toolchain>
+      </toolchains>`;
+
+    fs.mkdirSync(m2Dir, { recursive: true });
+    fs.writeFileSync(toolchainsFile, originalFile);
+    expect(fs.existsSync(m2Dir)).toBe(true);
+    expect(fs.existsSync(toolchainsFile)).toBe(true);
+
+    await toolchains.createToolchainsSettings({
+      jdkInfo,
+      settingsDirectory: m2Dir,
+      overwriteSettings: false
+    });
+
+    expect(fs.existsSync(m2Dir)).toBe(true);
+    expect(fs.existsSync(toolchainsFile)).toBe(true);
+    expect(fs.readFileSync(toolchainsFile, 'utf-8')).toEqual(originalFile);
+  }, 100000);
+
+  it('generates valid toolchains.xml with minimal configuration', () => {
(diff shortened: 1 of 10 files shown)


Okeanos:
I would very much appreciate if @dmitry-shibanov or someone else from the maintainers could chime in here and tell me what I have to do in order to get this merged. Thanks in advance.

laeubi:
@dmitry-shibanov can you help here? That would be very helpful!

src/toolchains.ts
```diff
@@ -0,0 +1,85 @@
...
+export function generate(version: string, vendor: string, jdkHome: string) {
+  const xmlObj: { [key: string]: any } = {
+    toolchains: {
+      '@xmlns': 'http://maven.apache.org/TOOLCHAINS/1.1.0',
+      '@xmlns:xsi': 'http://www.w3.org/2001/XMLSchema-instance',
+      '@xsi:schemaLocation':
+        'http://maven.apache.org/TOOLCHAINS/1.1.0 http://maven.apache.org/xsd/toolchains-1.1.0.xsd',
+      toolchain: [
+        {
+          type: 'jdk',
```
laeubi, line 47: Maybe it would be possible to have an "id" ([like described here](https://maven.apache.org/plugins/maven-toolchains-plugin/toolchains/jdk.html)) defined in the action as a parameter that is then used here? like
`id: '<id from action>'` t

laeubi (approved):
This is a great feature and would simplify setup for maven based builds!

src/toolchains.ts
Okeanos: @laeubi that should be possible now

laeubi: Man thanks, lets hope this get merged soon!

laeubi:
@marko-zivic-93 can you maybe review/ merge this?

Okeanos:
@laeubi there's an ongoing discussion in #276 where some folks from GitHub maintaining this repository chimed in as well. If you are in favour of this and maybe know some other people who'd like this, I suggest you post there and like the issue to give visible feedback to the maintainers.

Okeanos:
The failing validation checks were caused by connection timeouts during installation. Not by anything in this PR.

brcrista:
@Okeanos could you point us to a sample workflow run that uses the new functionality?  That would help me understand what we're adding here.

I saw the YAML snippet you added for the docs, but I'm also curious how the toolchain gets consumed in later steps.

laeubi:
You can take a look here:

https://github.com/eclipse-platform/eclipse.platform/blob/master/.github/workflows/build.yml

currently we maintain [a hand written toolchains.xml](https://github.com/eclipse-platform/eclipse.platform/blob/master/.github/toolchains.xml) and need to pass that to the maven command see:
https://github.com/eclipse-platform/eclipse.platform/blob/b3ad2b40185da67b124e4359468f5d00057776c9/.github/workflows/build.yml#L31

with an automatic one like proposed here, it would be just a matter of calling `mvn` without any extra parameter and always having the setup java (while currently the toolchains and java setup can easily get out of sync)

Okeanos:
@brcrista I setup a minimal project with pipelines that are hopefully self-descriptive here: https://github.com/Okeanos/setup-java-example/actions

It appears that Gradle benefits less from this PR on GitHub Actions than anticipated from [Gradle#14903](https://github.com/gradle/gradle/issues/14903) even for e.g. Gradle 7.3 which did not support Maven Toolchains detection.

Some jobs, such as the Java 8 and 11 are supposed to fail (technically, even for Gradle but it has auto-detection that works on GitHub Actions and for some reason matrix builds behave different from what I expected) because no valid JDK was found/supposed to be available/configured and the toolchains requirements in the `pom.xml` and `build.gradle` enforce an exact version requirement.

I also found and fixed a bug in my implementation and improved the tests as well. Asking for in-depth examples was a good idea ❤️ .

Additionally, I discovered that the `outputs.java-version` value from the `setup-java` task is actually incompatible with Toolchains declarations because it is the _exact_ version (as evident in the Job called `Java Default Actions sample` step `Show Toolchains (Debug)` whereas the Java version for major releases is required (as used in the `setup-java.inputs.java-version`. Even the `distribution` value is … not really good. I would have expected `temurin` to be the value not `Temurin-Hotspot`; `distribution` is special because we can filter by `vendor` within Maven builds and "Temurin-Hotspot" isn't a vendor … "Temurin" could be interpreted as one; technically correct would probably be `Eclipse Temurin` for this example (and is reported by Gradle as well).

If you have any further questions or need better examples (for an arbitrary value of "better") please let me know.

Okeanos:
Just got the pipeline notifications … gonna fix the formatting issues etc. Could have sworn I ran that formatting etc. after checking with the readme/contribution guidelines. Sorry about that :(

src/toolchains.ts
```diff
@@ -0,0 +1,133 @@
...
+import * as fs from 'fs';
+import * as os from 'os';
+
+import { create as xmlCreate } from 'xmlbuilder2';
+import * as constants from './constants';
+import { getBooleanInput } from './util';
+
+export async function configureToolchains(jdkHome: string) {
+  const version = core.getInput(constants.INPUT_JAVA_VERSION);
+  const vendor = core.getInput(constants.INPUT_DISTRIBUTION);
```
brcrista, line 14: We already read these in the `run()` function (in `setup-java.ts`).  Can you just pass them as arguments to this function?

Okeanos:
As an aside and something that was not yet fully addressed I believe in either the PR or the issue: we can make the toolchains generation optional/opt-in and properly document that in the readme.

This would ensure there are absolutely no breaking changes for existing users with working solutions and reduce the maintenance/support headache you rightfully pointed out – if they ever get around to upgrading (and reading about the changes in the update) they can still decide to switch. New users will hopefully see the notes and decide for themselves what they like best.

Any thoughts? Personally I am fine with either.

src/toolchains.ts
```diff
@@ -0,0 +1,133 @@
+import * as path from 'path';
+import * as core from '@actions/core';
+import * as io from '@actions/io';
+
+import * as fs from 'fs';
+import * as os from 'os';
+
+import { create as xmlCreate } from 'xmlbuilder2';
+import * as constants from './constants';
+import { getBooleanInput } from './util';
```
brcrista, line 10: Unfortunately we don't really have a consistent import style in this codebase.  Let's just go ahead and group them in a way where it's clear where to insert new ones.

```suggestion
import * as fs from 'fs';
import * as os from 'os';
import * as path from 'path';
import * as core from '@actions/core';
import * as io from '@actions/io';
import * as constants from './constants';

import { getBooleanInput } from './util';
import { create as xmlCreate } from 'xmlbuilder2';
```

src/toolchains.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export async function createToolchainsSettings(
+  version: string,
+  vendor: string,
+  id: string,
+  jdkHome: string,
+  settingsDirectory: string,
+  overwriteSettings: boolean
+) {
```
brcrista, line 37:
```suggestion
export async function createToolchainsSettings({
  version: string,
  vendor: string,
  id: string,
  jdkHome: string,
  settingsDirectory: string,
  overwriteSettings: boolean
}) {
```

Since this has a lot of parameters, it would be good to require an object with all the settings.  That way a `string` can't accidentally get passed in the wrong position.

brcrista:
There are some CI failures.  Running `npm format` and `npm build` should fix those.

src/toolchains.ts
Okeanos: Fixed

src/toolchains.ts
Okeanos: Fixed

src/toolchains.ts
Okeanos: Fixed

Okeanos:
> As an aside and something that was not yet fully addressed I believe in either the PR or the issue: we can make the toolchains generation optional/opt-in and properly document that in the readme.
> 
> This would ensure there are absolutely no breaking changes for existing users with working solutions and reduce the maintenance/support headache you rightfully pointed out – if they ever get around to upgrading (and reading about the changes in the update) they can still decide to switch. New users will hopefully see the notes and decide for themselves what they like best.
> 
> Any thoughts? Personally I am fine with either.

Along with review fixes I did some additional changes pertaining to this:

- rename the `toolchain-id` flag to `mvn-toolchain-id`
- add a `mvn-toolchain` boolean flag (currently set to `true` by default) allowing users to easily toggle generation/modification of the file per `setup-java` invocation
- expand the documentation the readme and advanced-usage docs somewhat to make usage and intentions clearer

laeubi:
@Okeanos mayn many thanks this looks really great!

brcrista:
> add a `mvn-toolchain` boolean flag (currently set to `true` by default) allowing users to easily toggle generation/modification of the file per `setup-java` invocation

I've been thinking about this and haven't come to a decision yet ... I'm leaning towards doing it though.  Since the intent is to avoid a potential breaking change, we would have to set it to `false` by default and then either switch it to `true` or remove the input on a major version upgrade.

Okeanos:
> > add a `mvn-toolchain` boolean flag (currently set to `true` by default) allowing users to easily toggle generation/modification of the file per `setup-java` invocation
> 
> I've been thinking about this and haven't come to a decision yet ... I'm leaning towards doing it though. Since the intent is to avoid a potential breaking change, we would have to set it to `false` by default and then either switch it to `true` or remove the input on a major version upgrade.

Take your time – I am in no rush and just happy this is considered for integration. Waiting another few days and fixing any outstanding issues in the meanwhile is not a problem for me and I'll gladly try to incorporate any changes you would like to have made.
Spontaneously, I am not even sure removing the `mvn-toolchain` boolean flag is ultimately necessary based on current integration complexity of this whole feature within the setup task vs. the inherit (perceived) benefit of additional flexibility it offers to consumers/users. Inverting the boolean flag – assuming `false` is the initial default – to push users towards a "better" future with less custom code to maintain on their end sounds like a good future step, though.

laeubi:
:+1: for `mvn-toolchain = true` as the default! I can't think of any reason doing this manually instead... so make it default and let the flag vanish on next major release instead.

src/setup-java.ts
```diff
@@ -16,6 +17,7 @@ async function run() {
     const jdkFile = core.getInput(constants.INPUT_JDK_FILE);
     const cache = core.getInput(constants.INPUT_CACHE);
     const checkLatest = getBooleanInput(constants.INPUT_CHECK_LATEST, false);
+    const mvnToolchain = getBooleanInput(constants.INPUT_MVN_TOOLCHAIN, true);
```
dmitry-shibanov, line 20: Should it be enabled by default ?

dmitry-shibanov:
Could you please sync with the main branch ?

Okeanos:
> Could you please sync with the main branch ?

As far as I can currently tell the branch builds on top of the current main branch. The GitHub UI also says "This branch is [1 commit ahead](https://github.com/Okeanos/setup-java/compare/actions:setup-java:main...maven-toolchains-support) of actions:main." the one commit being my commit introducing this feature.

If you introduce any new commits to main before this is merged I'll incorporate those changes of course.

Okeanos:
@brcrista @dmitry-shibanov I'll be switching to `default: false` soon based on the discussion above. Additionally, I noticed there are a couple of PRs such as #368 you are currently working on that will have some impact here. Please let me know when I can expect that to be merged and I'll update my PR to incorporate this.

src/constants.ts
```diff
@@ -20,3 +20,9 @@ export const INPUT_CACHE = 'cache';
 export const INPUT_JOB_STATUS = 'job-status';
 
 export const STATE_GPG_PRIVATE_KEY_FINGERPRINT = 'gpg-private-key-fingerprint';
+
+export const M2_DIR = '.m2';
+export const MVN_SETTINGS_FILE = 'settings.xml';
+export const MVN_TOOLCHAINS_FILE = 'toolchains.xml';
+export const INPUT_MVN_TOOLCHAIN = 'mvn-toolchain';
+export const INPUT_MVN_TOOLCHAIN_ID = 'mvn-toolchain-id';
```
brcrista, line 28: These new inputs need to be added to the [action.yml](https://github.com/actions/setup-java/blob/5b36705a13905facb447b6812d613a06a07e371d/action.yml#L5).

brcrista:
> I've been thinking about this and haven't come to a decision yet ... I'm leaning towards doing it though. Since the intent is to avoid a potential breaking change, we would have to set it to false by default and then either switch it to true or remove the input on a major version upgrade.

Ok, so I consider defaulting `mvn-toolchain` to `true` to be a breaking change because it will cause the toolchains file to be overwritten by side effect, and it alters the behavior of Maven.  

I think we can just remove the `mvn-toolchain` input and make the next release once this PR goes in a major-version release.

@dmitry-shibanov are you planning to merge https://github.com/actions/setup-java/pull/368 first?  It seems like this PR will need some reaction work to add all Java versions to the Maven toolchain file.

Okeanos:
> > I've been thinking about this and haven't come to a decision yet ... I'm leaning towards doing it though. Since the intent is to avoid a potential breaking change, we would have to set it to false by default and then either switch it to true or remove the input on a major version upgrade.
> 
> Ok, so I consider defaulting `mvn-toolchain` to `true` to be a breaking change because it will cause the toolchains file to be overwritten by side effect, and it alters the behavior of Maven.
> 
> I think we can just remove the `mvn-toolchain` input and make the next release once this PR goes in a major-version release.
> 
> @dmitry-shibanov are you planning to merge #368 first? It seems like this PR will need some reaction work to add all Java versions to the Maven toolchain file.

In that case I'll wait for #368 to be merged and make the necessary changes for that alongside removing the `mvn-toolchains` input + adding `mvn-toolchain-id` to the `action.yml`.

kingtaxi51:
Dam I don't understand I'm experimenting with intragram trying to make
money I saw a web site Indian mart where I can sell other products I know I
can sell anything I'm very good at it I want to make a platform that I can
promote different types of things to sell them and make money with comision
I'm still learning how to tag and a few things but I made alot of money
with Facebook selling cars so I know online traffic is the best and selling
something with even having it is better I have three kids men I'm 42 years
old trying to find a better way to take care of my family

Okeanos:
@brcrista @dmitry-shibanov I updated the PR for setup-java to support Toolchains now that #368 (install multiple JDKs at once) has been merged.

Based on the previous findings:

- `actions.yml` documentation added
- `mvn-toolchain` boolean flag removed
- formatting etc. fixed (🤞 )

Compared to the previous iteration I did some additional changes for better compatibility with exiting setup options:

- users can now specify a `mvn-toolchain-vendor` to override the `distribution` input value that was previously used at all times. A useful example as to why this is necessary is included in the Advanced Usage documentation based on the `jdkfile`-self-supplied-JDK-option.
- `mvn-toolchain-vendor` can only be specified one time and effectively is a fancy name for `distribution` for multi-line input values of `java-version`
- `mvn-toolchain-id` now accepts multi-line input and will defer to `java-versions`, i.e. either a specific ID is given for every java version specified or the id value(s) are ignored completely. Fewer edge cases this way.

If there is anything else you would like me to change, let me know and I'll have a look.

__tests__/toolchains.test.ts
```diff
@@ -0,0 +1,249 @@
+import io = require('@actions/io');
+import fs = require('fs');
+import path = require('path');
```
brcrista, line 3: Could you change this to the `import * as ...` syntax?  That avoids some subtle gotchas

src/toolchains.ts
```diff
@@ -0,0 +1,143 @@
...
+import * as constants from './constants';
+
+import { getBooleanInput } from './util';
+import { create as xmlCreate } from 'xmlbuilder2';
+
+export async function configureToolchains(
+  version: string,
+  distributionName: string,
+  jdkHome: string,
+  toolchainId: string | undefined
```
brcrista, line 15:
```suggestion
  toolchainId?: string
```

TypeScript has sugar for optional parameters

brcrista (approved):
A couple minor comments, but nothing blocking merge.  @dmitry-shibanov could you or one of your teammates take a look also?

__tests__/toolchains.test.ts
Okeanos: Doing that will throw "Cannot redefine property" errors because of the `spyOSHomedir = jest.spyOn(os, 'homedir')` in the test setup. Interestingly it works for `core` (where we do the same thing).
I am not well versed enough in either TypeScript or Jest to propose a better alternative. Any suggestions?

src/toolchains.ts
```diff
@@ -0,0 +1,143 @@
...
+  vendor,
+  id,
+  jdkHome,
+  settingsDirectory,
+  overwriteSettings
+}: {
+  version: string;
+  vendor: string;
+  id: string;
+  jdkHome: string;
```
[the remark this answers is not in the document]
Okeanos: Implemented

src/toolchains.ts
Okeanos: Fixed

__tests__/toolchains.test.ts
Okeanos: I went ahead and fixed this for all except `os` for now.

src/toolchains.ts
```diff
@@ -0,0 +1,154 @@
...
+  if (fs.existsSync(location)) {
+    return fs.readFileSync(location, {
+      encoding: 'utf-8',
+      flag: 'r'
+    });
+  }
+  return '';
+}
+
+async function write(directory: string, settings: string, overwriteSettings: boolean) {
```
e-korolevskii, line 136: I think the function name should be clearer

src/toolchains.ts
```diff
@@ -0,0 +1,154 @@
...
+  return xmlObj.end({
+    format: 'xml',
+    wellFormed: false,
+    headless: false,
+    prettyPrint: true,
+    width: 80
+  });
+}
+
+async function readExisting(directory: string) {
```
e-korolevskii, line 125: I think the function name should be clearer

src/toolchains.ts
```diff
@@ -0,0 +1,154 @@
...
+    jdkInfo.version,
+    jdkInfo.vendor,
+    jdkInfo.id,
+    jdkInfo.jdkHome
+  );
+  await write(settingsDirectory, updatedToolchains, overwriteSettings);
+}
+
+// only exported for testing purposes
+export function generate(
```
e-korolevskii, line 69: I think the function name should be clearer

src/toolchains.ts
```diff
@@ -0,0 +1,154 @@
...
+  original: string,
+  version: string,
+  vendor: string,
+  id: string,
+  jdkHome: string
+) {
+  let xmlObj;
+  if (original && original.length > 0) {
```
e-korolevskii, line 77:
```suggestion
  if (original?.length) {
```

vsafonkin (approved):

laeubi (approved):
Go for it!

e-korolevskii (approved):

panticmilos (approved):

IvanZosimov (approved):

src/toolchains.ts
Okeanos: Changed to `readExistingToolchainsFile`

src/toolchains.ts
Okeanos: Fixed.

src/toolchains.ts
Okeanos: Changed to `writeToolchainsFileToDisk`.

marko-zivic-93 (approved):

brcrista:
Thank you @Okeanos !

PR #284, opened 2022-01-21
Add cacerts parameter, which can copy existing cacerts into JDK

mdvorak:
**Description:**
In enterprise environment, custom CA truststore is often used. This parameter allows automatic copy of `cacerts` file into new Java installation.

**Check list:**

There are no e2e test yet, I can add some if you are interested in merging this.

Note that this modifies actual JDK directory. I was worried about tool cache, but it seems that tool is cached (copied) right after extraction, so copying this file into final directory does not modify tool cache at all. It should be safe.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,10 +20,10 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - uses: actions/checkout@v2
-      - name: Setup Node.JS 12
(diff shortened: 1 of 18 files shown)


brunoborges:
The `cacerts` idea may still be useful, but this branch is too outdated to merge safely. Recommendation: decide whether to pursue via a new scoped PR on current `main`.

brunoborges:
Closing as outdated due to age and merge conflicts. Continuing this proposal in #1035.

PR #290, opened 2022-02-07
Update Default runtime to node16

thboop:
Node 12 has an end of life on April 30, 2022.

This PR updates the default runtime to [node16](https://github.blog/changelog/2021-12-10-github-actions-github-hosted-runners-now-run-node-js-16-by-default/), rather then node12. 

This is supported on all Actions Runners v2.285.0 or later.

**Description:**
Describe your changes.

**Related issue:**

**Check list:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,10 +20,10 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - uses: actions/checkout@v2
-      - name: Setup Node.JS 12
+      - name: Setup Node.JS 16
         uses: actions/setup-node@v2
         with:
-          node-version: 12.x
+          node-version: 16.x
           cache: npm
       - run: npm ci
       - run: npm run build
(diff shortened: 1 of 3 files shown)


dmitry-shibanov (approved):

MaksimZhukov (approved):

jglick:
https://github.com/actions/setup-java/releases/tag/v3.0.0 lists this as a “breaking change” but does not really explain how you would know whether or not you need to pay attention. For someone encountering this update with no further context in a Dependabot bump of the action (https://github.com/jenkinsci/cloudbees-folder-plugin/pull/217 for example), it is hard to tell whether or not it would be safe to apply.

> This is supported on all Actions Runners v2.285.0 or later.

implies that workflows using GitHub-supplied runners would be unaffected and only people using self-hosted runners need care, but I am just guessing here.

965022-oss:
Merge #290 into actions/starter-workflows, 3a95482282/f4dbb8424c,
label: Update default runtime to node 16 @965022-oss

Issue #295, opened 2022-02-28
Is v3 officially released?


artamonovkirill:
**Description:**
I got a PR from Renovate bot bumping the `setup-java` action to v3.

**Task version:**
v3:
```
     - uses: actions/setup-java@v3
        with:
          java-version: 17
          distribution: adopt
          cache: gradle
```

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Repro steps:**  
https://github.com/artamonovkirill/ninshubur/pull/189

**Expected behavior:**
CI step succeeds

**Actual behavior:**
CI steps fails due to out-of-range error:
```
Run actions/setup-java@v3
  with:
    java-version: 17
    distribution: adopt
    cache: gradle
    java-package: jdk
    architecture: x64
    check-latest: false
    server-id: github
    server-username: GITHUB_ACTOR
    server-password: GITHUB_TOKEN
    overwrite-settings: true
    job-status: success
Trying to resolve the latest version from remote
Resolved latest version as 17.0.[2](https://github.com/artamonovkirill/ninshubur/runs/5359803125?check_suite_focus=true#step:6:2)+8
Trying to download...
Downloading Java 17.0.2+8 (Adopt-Hotspot) from https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.2%2B8/OpenJDK17U-jdk_x64_linux_hotspot_17.0.2_8.tar.gz ...
Extracting Java archive...
/usr/bin/tar xz --warning=no-unknown-keyword -C /home/runner/work/_temp/db9051e8-02a1-4d70-a166-91e[3](https://github.com/artamonovkirill/ninshubur/runs/5359803125?check_suite_focus=true#step:6:3)325db[4](https://github.com/artamonovkirill/ninshubur/runs/5359803125?check_suite_focus=true#step:6:4)[5](https://github.com/artamonovkirill/ninshubur/runs/5359803125?check_suite_focus=true#step:6:5)0 -f /home/runner/work/_temp/03d793fc-d48f-4a35-ba8c-beb8[6](https://github.com/artamonovkirill/ninshubur/runs/5359803125?check_suite_focus=true#step:6:6)95bfd11
[...]

PR #297, opened 2022-03-01
Improve error handling for restore caching

dmitry-shibanov:
**Description:**
In scope of this pull request we add error handling to throw error only for ValidationError.

**Related issue:** https://github.com/actions/setup-java/issues/295

**Check list:**


ChrisCarini (approved):

Pompo17102:
Deposit through the bitcoin wallet for any help 
     bc1q5v34g2dym8c7x5mpfp7wdqp5hse38p99p4dggx

brcrista (approved):
I agree with the change, but how does it relate to https://github.com/actions/setup-java/issues/295?

brcrista:
@actions/actions-cache, do you have any feedback?  For context, the `setup-` actions use the `@actions/cache` package to cache dependencies directly, which lets you omit `actions/cache` as a separate step.

dmitry-shibanov:
Hello @brcrista. Actually, as I checked the issue reproduces with node 16 only. On the previous action version it works as expected. I think it's related itself to the change of the default node version and buffer size. I could reproduce the same behaviour with action/cache when I removed [this error handling logic](https://github.com/actions/cache/blob/main/src/restore.ts#L59-L64) and archive size was more than 2147483647.
 - https://github.com/actions/toolkit/blob/39b96406423e444b058d2602e2cea8f889342ee4/packages/cache/src/internal/downloadUtils.ts#L113-L115

aparna-ravindra:
might this be related to https://github.com/actions/toolkit/pull/1030 ?

dmitry-shibanov:
Thank you @aparna-ravindra. It works. I've updated the pull request.

dmitry-shibanov:
I'm going to close the pull request because the issue was resolved in scope of this pull request https://github.com/actions/setup-java/pull/308

PR #327, opened 2022-05-06
Updates the sbt cache docs in the README

adamnfish:
**Description:**

The README's description of sbt's cache behaviour does not match the implementation. An `sbt` build can be modified from multiple places and the current cache invalidation behaviour correctly captures these cases. The README only mentions `build.sbt` files, which might mean teams with more experience using `sbt` unnecessarily add their own cache patterns here.

Here's the (current) implementation, for reference:
https://github.com/actions/setup-java/blob/860f60056505705214d223b91ed7a30f173f6142/src/cache.ts#L43

**Related issue:**
N/A

**Check list:**


fmeriaux (approved):
LGTM 👍

Good point, I had forgotten to edit this section in my contribution.

marko-zivic-93 (approved):

dmitry-shibanov (approved):

Issue #359, opened 2022-08-04
Allow custom target directory for exploding `jdkFile` into


sormuras:
**Description:**
Introduce an optional input for specifying a custom target directory for exploding a `jdkFile` into.

**Justification:**
Installing multiple JDKs that share the same JDK feature release number (like `19`) is not possible at the moment.

For example, after downloading two JDK archives that both are based on JDK 19, the second run of `setup-java@v3` detects that a Java 19 is already installed and uses it; instead of extracting the second JDK archive into a new and different directory:

```yml
    - name: 'Install custom JDK 19'
      uses: actions/setup-java@v3
      with:
        distribution: jdkfile
        jdkFile: .../openjdk-19-jextract+2-3_linux-x64_bin.tar.gz
        java-version: 19.0.0+0-jextract

    - name: 'Install full JDK 19'
      uses: actions/setup-java@v3
      with:
        distribution: jdkfile
        jdkFile: .../openjdk-19-ea+33_linux-x64_bin.tar.gz
        java-version: 19
```

The output of the second run reads like:

```yml
Run actions/setup-java@v3
  with:
    java-version: 19
    distribution: jdkfile
    jdkFile: .../openjdk-19-ea+33_linux-x64_bin.tar.gz
    java-package: jdk
    architecture: x64
    check-latest: false
    server-id: github
    server-username: GITHUB_ACTOR
    server-password: GITHUB_TOKEN
    overwrite-settings: true
    job-status: success
  env:
    JAVA_HOME: /opt/hostedtoolcache/Java_jdkfile_jdk/19.0.0-0-jextract/x64
Resolved Java 19.0.0+0-jextract from tool-cache
Setting Java 19.0.0+0-jextract as default
```

**Expectation:**

With a custom target directory specified for each run (`openjdk-19-jextract+2-3` and `openjdk-19-ea+33`), there would be two dedicated JDK installations. Like:

- /opt/hostedtoolcache/Java_jdkfile_jdk/openjdk-19-jextract+2-3/x64
- /opt/hostedtoolcache/Java_jdkfile_jdk/openjdk-19-ea+33/x64

**Work-around:**
Pass fake values to `java-version` in order to generated dedicated target directories. See https://github.com/oracle-actions/setup-java/issues/34 for details.

denismakogon:
[...]

PR #363, opened 2022-08-10
Installing multiple JDKs that share the same JDK feature release number

dsame:
**Description:**
 Instaling JDK with same feature version requires to set either [artificial](https://github.com/actions/setup-java/issues/359#issuecomment-1206281095) or non-semver valid `java-version` input.
 
 Samples of such JDKs:
  - openjdk-19-ea+33_linux-x64_bin.tar.gz
  - openjdk-19-jextract+2-3_linux-x64_bin.tar.gz

Currently it is not possible to set input to something like `19-ea` and users of action has to think out 'semver-like' java-version inout that is not straightforward. The PR relaxed the requirement for local distributons.

**Related issue:**
[Add link to the related issue.](https://github.com/actions/setup-java/issues/359)

**Check list:**

### DIFF
--- a/src/distributions/base-installer.ts
+++ b/src/distributions/base-installer.ts
@@ -129,7 +129,7 @@ export abstract class JavaBase {
       stable = false;
     }
 
-    if (!semver.validRange(version)) {
+    if (this.distribution !== 'jdkfile' && !semver.validRange(version)) {
       throw new Error(
         `The string '${version}' is not valid SemVer notation for a Java version. Please check README file for code snippets and more detailed information`
       );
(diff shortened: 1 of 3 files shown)


marko-zivic-93 (approved):

src/distributions/base-installer.ts
```diff
@@ -129,7 +129,7 @@ export abstract class JavaBase {
       stable = false;
     }
 
-    if (!semver.validRange(version)) {
+    if (this.distribution !== 'jdkfile' && !semver.validRange(version)) {
```
dmitry-shibanov, line 132: If it does not follow SemVer notation it will lead possible drawback to versions in toolcache directory because toolkit find method relies on SemVer notation.
- https://github.com/actions/toolkit/blob/main/packages/tool-cache/src/tool-cache.ts#L495

PR #383, opened 2022-09-19
Add versions-manifest.json for Microsoft Build of OpenJDK

dmitry-shibanov:
**Description:**
In scope of this pull request we change behaviour for Microsoft Build of OpenJDK. We add json containing versions which can be installed. 
**Related issue:**

**Check list:**


versions-manifest.json
```diff
@@ -0,0 +1,181 @@
+[
```
maxim-lobanov, line 1: I think this file should be renamed and placed to some subfolder rather than root of repository. Or placed into separate repo `actions/java-versions`

marko-zivic-93 (approved):

PR #397, opened 2022-10-28
always check postfix "Contents/Home" on macOS

### LINKED ISSUE #396: JAVA_HOME is not set correctly on MacOS for local (jdkFile) installed JDKs that are present in the tool cache
**Description:**
Local installed JDKs using the jdkFile option that are already present in the tool cache are not checked for suffix 'Contents/Home' on macos, so as a result the JAVA_HOME is not set correctly.

**Task version:**
v3

**Platform:**
- [ ] Ubuntu
- [x] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [x] Self-hosted

**Repro steps:**  
- run a workflow with setup-java with a jdkFile option on a self-hosted runner on MacOS using a JDK that places files in 'Contents/Home'
- the first time this works fine. JAVA_HOME is set correctly including 'Contents/Home'
- run the workflow again, this time the JDK will already be present in the tool cache
- setup-java now sets JAVA_HOME without the 'Contents/Home' suffix

**Expected behavior:**
When the JDK is available in the tool cache, JAVA_HOME must be set correctly.

**Actual behavior:**
When the JDK is available in the tool cache, JAVA_HOME is set without checking if the JDK is placed in 'Contents/Home'

erwin1:
**Description:**
Also check if the JDK is placed in 'Contents/Home' on macos when it's found in the tool cache.

**Related issue:**
Fixes #396 

**Check list:**


IvanZosimov (approved):

IvanZosimov:
Hi, @erwin1 👋 Thank you for the PR, I've synced PR with the current main, linted and rebuilded solution.

dmitry-shibanov (approved):

MaksimZhukov (approved):

PR #401, opened 2022-11-04
Add support for Oracle JDK

fniephaus:
**Description:**
This PR adds Oracle JDK to the list of supported distributions.

**Related issue:** https://github.com/actions/setup-java/issues/69

**Check list:**

### DIFF
--- a/src/distributions/oracle/installer.ts
+++ b/src/distributions/oracle/installer.ts
@@ -1,0 +1,102 @@
+import * as core from '@actions/core';
+import * as tc from '@actions/tool-cache';
+
+import fs from 'fs';
+import path from 'path';
+
+import { JavaBase } from '../base-installer';
+import { JavaDownloadRelease, JavaInstallerOptions, JavaInstallerResults } from '../base-models';
+import { extractJdkFile, getDownloadArchiveExtension } from '../../util';
+import { HttpCodes } from '@actions/http-client';
+
+const ORACLE_DL_BASE = 'https://download.oracle.com/java'
+
+export class OracleDistribution extends JavaBase {
+  constructor(installerOptions: JavaInstallerOptions) {
+    super('Oracle', installerOptions);
+  }
+
+  protected async downloadTool(javaRelease: JavaDownloadRelease): Promise<JavaInstallerResults> {
+    core.info(
+      `Downloading Java ${javaRelease.version} (${this.distribution}) from ${javaRelease.url} ...`
+    );
+    const javaArchivePath = await tc.downloadTool(javaRelease.url);
+
+    core.info(`Extracting Java archive...`);
+    let extension = getDownloadArchiveExtension();
+
+    let extractedJavaPath = await extractJdkFile(javaArchivePath, extension);
+
+    const archiveName = fs.readdirSync(extractedJavaPath)[0];
+    const archivePath = path.join(extractedJavaPath, archiveName);
+    const version = this.getToolcacheVersionName(javaRelease.version);
+
+    let javaPath = await tc.cacheDir(
+      archivePath,
+      this.toolcacheFolderName,
+      version,
+      this.architecture
+    );
+
+    return { version: javaRelease.version, path: javaPath };
+  }
+
+  protected async findPackageForDownload(range: string): Promise<JavaDownloadRelease> {
+    const arch = this.distributionArchitecture();
+    if (arch !== 'x64' && arch !== 'aarch64') {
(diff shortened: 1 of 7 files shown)


brunoborges (approved):
LGTM

fniephaus:
Thanks for approving, @brunoborges. I just pushed a commit that should make the tests work on macOS and Windows.

src/distributions/oracle/installer.ts
```diff
@@ -0,0 +1,103 @@
...
+      version,
+      this.architecture
+    );
+
+    return { version: javaRelease.version, path: javaPath };
+  }
+
+  protected async findPackageForDownload(range: string): Promise<JavaDownloadRelease> {
+    const arch = this.distributionArchitecture();
+    if (arch !== 'x64' && arch !== 'aarch64') {
```
dmitry-shibanov, line 46: I think Oralce provides java binaries for `aarch64`.

src/distributions/oracle/installer.ts
```diff
@@ -0,0 +1,103 @@
...
+      major = range.split('.')[0];
+      fileUrl = `${ORACLE_DL_BASE}/${major}/archive/jdk-${range}_${platform}-${arch}_bin.${extension}`;
+    } else {
+      major = range;
+      fileUrl = `${ORACLE_DL_BASE}/${range}/latest/jdk-${range}_${platform}-${arch}_bin.${extension}`;
+    }
+
+    if (parseInt(major) < 17) {
+      throw new Error('Oracle JDK is only supported for JDK 17 and later');
+    }
```
dmitry-shibanov, line 72: I think we can remove and add it to the documentation because if the versions is not found it will return 404 code. It can be handled further code checks.

dmitry-shibanov:
Hello @fniephaus. Could you please update documentation for advanced configuration with examples for Oracle JDK ? Besides, If I understood correctly, it is impossible to grab all versions from API ?

src/distributions/oracle/installer.ts
sormuras: Yes. Oracle does provide `x64` and `aarch64` binaries for Linux and macOS. That's why this check throws an error for non-`x64` and non-`aarch64` architectures.

Note: no `aarch64` binaries for Windows, though https://www.oracle.com/java/technologies/downloads/

src/distributions/oracle/installer.ts
sormuras: Adding this limitation to the documentation is a good point.

Keeping it here has also an additional value, as users passing lower values for Java's "major" version will see an explicit error message that this vendor only supports "JDK 17 and later".

src/distributions/oracle/installer.ts
fniephaus: Related part in the tests: https://github.com/actions/setup-java/pull/401/files#diff-14331bf7630250eee172ff240c74a608009a7580b009948f84e3af1ea00f328eR76

dmitry-shibanov:
> > Could you please update documentation for advanced configuration with examples for Oracle JDK ?
> 
> I'll add an example to https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md
> 
> > Besides, If I understood correctly, it is impossible to grab all versions from API ?
> 
> What do you mean with "grab all version from API" ?
> 
> Find a list and a description of supported versions at https://www.oracle.com/java/technologies/jdk-script-friendly-urls/

For now the logic constructs url from major or full version. Is it possible to get all available versions ?

sormuras:
Ah, understood. AFAIK, it's not possible to get a list of all binaries via an API call.

There is however a listing available per "major" version at: https://www.oracle.com/java/technologies/downloads/archive/
- `19` ➡️ https://www.oracle.com/java/technologies/javase/jdk18-archive-downloads.html
- `18` ➡️ https://www.oracle.com/java/technologies/javase/jdk18-archive-downloads.html
- `17` ➡️ https://www.oracle.com/java/technologies/javase/jdk17-archive-downloads.html

sormuras (approved):
LGTM

panticmilos (approved):

dmitry-shibanov:
Thank you for your description @sormuras. In this case I'm going to merge the pull request.

fniephaus:
Thanks for merging this, @dmitry-shibanov. When can we expect a new release of this action so that users can use Oracle JDK and `actions/setup-java@v3` as recommended?

finaroux:
Hey I thing this tag was a mistake I have nothing to do with this tag (wrong person). Can you fix it please? Thanks!

PR #428, opened 2022-12-12
refactor: Use early return pattern to avoid nested conditions

jongwooo:
Signed-off-by: jongwooo <jongwooo.han@gmail.com>
Issue #420 

## Description

Return early is the way of writing functions or methods so that the expected positive result is returned at the end of the function and the rest of the code terminates the execution (by returning or throwing an exception) when conditions are not met.

See [actions/cache#1012](https://github.com/actions/cache/issues/1012), [actions/setup-node#639](https://github.com/actions/setup-node/pull/639)

In `util.ts`:

**AS-IS**

```typescript
export function isCacheFeatureAvailable(): boolean {
  if (!cache.isFeatureAvailable()) {
    if (isGhes()) {
      throw new Error(
        'Caching is only supported on GHES version >= 3.5...'
      );
    } else {
      core.warning('The runner was not able to contact...');
    }

    return false;
  }

  return true;
}
```

**TO-BE**

```typescript
export function isCacheFeatureAvailable(): boolean {
  if (cache.isFeatureAvailable()) {
    return true;
  }

  if (isGhes()) {
    throw new Error(
        'Caching is only supported on GHES version >= 3.5...'
    );
  }

  core.warning('The runner was not able to contact...');
  return false;
}
```

**Check list:**

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -85,17 +85,16 @@ export function isGhes(): boolean {
 }
 
 export function isCacheFeatureAvailable(): boolean {
-  if (!cache.isFeatureAvailable()) {
-    if (isGhes()) {
-      throw new Error(
-        'Caching is only supported on GHES version >= 3.5. If you are on a version >= 3.5, please check with your GHES admin if the Actions cache service is enabled or not.'
-      );
-    } else {
-      core.warning('The runner was not able to contact the cache service. Caching will be skipped');
-    }
-
-    return false;
+  if (cache.isFeatureAvailable()) {
+    return true;
   }
 
-  return true;
+  if (isGhes()) {
+    throw new Error(
+      'Caching is only supported on GHES version >= 3.5. If you are on a version >= 3.5, please check with your GHES admin if the Actions cache service is enabled or not.'
+    );
+  }
+
+  core.warning('The runner was not able to contact the cache service. Caching will be skipped');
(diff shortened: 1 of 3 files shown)


Xraduu (approved):

panticmilos:
hi @jongwooo, thank you for the PR. We will take a look at it

src/util.ts
```diff
@@ -85,17 +85,16 @@ export function isGhes(): boolean {
...
-    }
+  if (cache.isFeatureAvailable()) {
+    return true;
+  }
 
-    return false;
+  if (isGhes()) {
+    throw new Error(
+      'Caching is only supported on GHES version >= 3.5. If you are on a version >= 3.5, please check with your GHES admin if the Actions cache service is enabled or not.'
+    );
```
jongwooo, line 95: How about replace throw with log the warning here and return `false`?

panticmilos: Yes, let's apply the same updates we did on setup-go as well :)

jongwooo: @panticmilos Thank you for your review. I pushed a commit.

e-korolevskii (approved):

MaksimZhukov (approved):

dmitry-shibanov (approved):

PR #433, opened 2022-12-19
Update action to use reusable workflows

IvanZosimov:
**Description:**
In the scope of this PR we update workflows in the repo to use reusable-workflows from the [repo](https://github.com/actions/reusable-workflows).

**Related issue:**
[#4595](https://github.com/actions/runner-images-internal/issues/4595)

**Check list:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ on:
       - '**.md'
 
 jobs:
-  call-basic-validation:
+  call-basic-validation: # The reusable workflow can be found here: https://github.com/actions/reusable-workflows/blob/main/.github/workflows/basic-validation.yml
     name: Basic validation
     #uses: actions/reusable-workflows/.github/workflows/basic-validation.yml@main
     uses: IvanZosimov/reusable-workflows/.github/workflows/basic-validation.yml@main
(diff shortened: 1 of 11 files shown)


.github/workflows/build.yml
```diff
@@ -12,20 +12,6 @@ on:
...
-          node-version: 16.x
-          cache: npm
-      - run: npm ci
-      - run: npm run build
-      - run: npm run format-check
-      - run: npm test
\ No newline at end of file
+  call-basic-validation: # The reusable workflow can be found here: https://github.com/actions/reusable-workflows/blob/main/.github/workflows/basic-validation.yml
+    name: Basic validation
+    uses: actions/reusable-workflows/.github/workflows/basic-validation.yml@main
```
MaksimZhukov, line 17: I think we don't need such comments because it's obvious where the reusable workflow file is, judging by the value of the `uses` key

package.json
```diff
@@ -8,6 +8,7 @@
     "build": "ncc build -o dist/setup src/setup-java.ts && ncc build -o dist/cleanup src/cleanup-java.ts",
     "format": "prettier --write \"{,!(node_modules)/**/}*.ts\"",
     "format-check": "prettier --check \"{,!(node_modules)/**/}*.ts\"",
+    "lint": "",
```
MaksimZhukov, line 11: I think we should clarify why we need this
```suggestion
    "lint": "echo \"Fake command that does nothing. It is used in reusable workflows\"",
```

.github/workflows/build.yml
MaksimZhukov, line 17: Let's rename this workflow in accordance with the name of reusable workflow

dmitry-shibanov (approved):

e-korolevskii (approved):

marko-zivic-93 (approved):

PR #434, opened 2022-12-22
add support for artifact proxy repositories

kacamific:
**Description:**
This change supports teams that have configured their self-managed runners without internet access. They are proxying all the necessary public repositories via platforms like JFrog, Nexus etc. This can support especially teams that are using GHES.

I am proposing to add three inputs to action.yml
1- base url of the platform that is proxying the remote repositories 
2- the download link that needs to be rewritten so the java binaries are download from this platforms 
3- a context which might be necessary to build a proper download link. This might be necessary because the URL of the metadata.json file is different from the URL of the download link and the assumption is that teams have only one platform to proxy the remote repositores.

**Related issue:**
-

**Check list:**
- [ x] Mark if documentation changes are required.
- [ x] Mark if tests were added or updated to cover the changes.


kuhlman-labs:
@marko-zivic-93 Any feedback on this PR?

IvanZosimov:
Hi, @kuhlman-labs 👋 We will investigate it and get back with our feedback.

PR #441, opened 2023-01-08
feat: added support to pom.xml and build.gradle

augustomelo:
**Description:**
The idea of this PR is to extend the java-version-file to support pom.xml, because for now it only supports the java-version-file created by jEnv

**Related issue:**

**Check list:**


augustomelo:
This PR is still pending:

[ ] - Add unit e2e test to cover the other scenarios 
[ ] - Add documentation

panticmilos:
hi @augustomelo thank you for the PR, we will take a look at it

augustomelo:
Hello @panticmilos could you approve the workflow run ? I tested it locally with [act](https://github.com/nektos/act) and it was fine, just wanted to sure that I didn't break anything

augustomelo:
Hey, regarding the failing workflows, I believe there is something strange with them, because for example the code scanning:

* [GitHub Code Scanning / CodeQL](https://github.com/actions/setup-java/pull/441/checks?check_run_id=10718353077) complains in a file that is generated `dist/cleanup/index.js`
  * 97022: `if (util_1.isObject(p1) || (util_1.isString(p1) && (/^\s*</.test(p1) || /^\s*[\{\[]/.test(p1) || /^(\s*|(#.*)|(%.*))*---/.test(p1)))) {`
  * 97795: `else if (p1 !== null && /^(\s*|(#.*)|(%.*))*---/.test(p1))`

I believe this is coming from the  xmlbuilder2 when I create the object from the string [(here)](https://github.com/actions/setup-java/blob/ef956c675fdca09c71cc2115ce0705707398ae54/src/util.ts#L162), should I change the library ? Because I don't think I have any control over it.

The other workflows that failed are:
* [Basic validation / build (ubuntu-latest)](https://github.com/actions/setup-java/actions/runs/3941287822/jobs/6754396529#logs)
* [Basic validation / build (windows-latest)](https://github.com/actions/setup-java/actions/runs/3941287822/jobs/6754396612#logs)
* [Basic validation / build (macos-latest)](https://github.com/actions/setup-java/actions/runs/3941287822/jobs/6754396703#logs)
* [Licensed / Check licenses](https://github.com/actions/setup-java/actions/runs/3941287825/jobs/6754394481#logs)
* [Check dist/ / check-dist](https://github.com/actions/setup-java/actions/runs/3941287827/jobs/6754396049#logs)

Which are failing during the install dependencies step, it could be because the dependency that I added: `xpath` which has the same license as the one `xmlbuilder2`  (MIT).

Not sure how to proceed now, @panticmilos do you have any ideas here ?

augustomelo:
Hey @panticmilos  thanks for the reply! Sure thing I will have a look on it once I have the time 😄

tim-we:
Hey @augustomelo, first of all thanks for the PR, if this didn't exist I would have created my own ;)

But I wonder why you chose to parse the version via regex as that approach obviously limited. This project already has `xmlbuilder2` as a dependency which can also be used for parsing XML. For example we could do this:

```js
const doc = create(pomContentAsAString);
const mvnVersionNode = doc.find(n => n.node.nodeName === "maven.compiler.source", false, true);
const sourceVersion = mvnVersionNode.first().toString();
```

This way we can have a more robust implementation and support using the version specified in `maven-compiler-plugin`.

Are you still willing to work on this @augustomelo? Otherwise I can create a new PR. 

@panticmilos has this feature request been discussed? Are you willing to merge something like this?

augustomelo:
hey @tim-we, thanks for the review and I much appreciate the suggestion the solution is much better now! 

I am not sure why, but when I read the xmlbuilder2 I understood that it was not possible to create the xml object from a string, therefore I used the xpath and then moved to using regex.

tim-we:
Hey @augustomelo, I made [a PR to your fork](https://github.com/augustomelo/setup-java/pull/1)  to add support for versioning via `maven-compiler-plugin`. Please have a look. There are also some minor things we should address before this gets merged, such as variable names not being accurate anymore after the refactor. If this progresses any further I'd might do another code review.

src/util.ts
```diff
@@ -133,6 +145,107 @@ export function getVersionFromFileContent(
...
+    if (version !== null) {
+      return version;
+    }
+  }
+
+  return null;
+}
+
+function getByMavenProperties(xmlDoc: XMLBuilder): string | null {
+  const possibleTagsRegex = [
```
tim-we, line 180: These are not regular expressions, just tags. So please update the variable name. I suggest `versionTags`.

src/util.ts
```diff
@@ -133,6 +145,107 @@ export function getVersionFromFileContent(
...
+    'maven.compiler.source',
+    'maven.compiler.release',
+  ];
+
+  for (var tag of possibleTagsRegex) {
```
tim-we, line 185: Please use `const` instead of `var`.

src/util.ts
```diff
@@ -101,13 +103,23 @@ export function isCacheFeatureAvailable(): boolean {
 }
 
 export function getVersionFromFileContent(
+  fileName: string,
```
tim-we, line 106: Since the behavior of this function depends on the filename I suggest changing the name to `getVersionFromFile`.

src/util.ts
```diff
@@ -101,13 +103,23 @@ export function isCacheFeatureAvailable(): boolean {
...
   content: string,
   distributionName: string
 ): string | null {
-  const javaVersionRegExp = /(?<version>(?<=(^|\s|\-))(\d+\S*))(\s|$)/;
-  const fileContent = content.match(javaVersionRegExp)?.groups?.version
-    ? (content.match(javaVersionRegExp)?.groups?.version as string)
-    : '';
+  let fileContent = null;
```
tim-we, line 110: I suppose you wanted to keep the amount of changes minimal but I think the name `fileContent` is now slightly misleading, so I suggest something like `parsedVersion` instead.

augustomelo:
hey @tim-we  I didn't forgot about this, when I had the time I will apply the changes. I was thinking about also add support to gradle, but I currently don't know how to achieve that besides using RegEx, do you have any ideas ?

augustomelo:
hey @tim-we  refactored the code to align with your comments, feel free to resolve the conversation.

Added the support to builld. gradle using regex.

I didn't do the merge, because I am not sure if this PR is going to be merged

tim-we (approved):

augustomelo:
hey @tim-we  I had to do some changes on the code since the testes [were failing in correctly identifying the java version](https://github.com/augustomelo/setup-java/actions/runs/8866172595), [now](https://github.com/augustomelo/setup-java/actions/runs/8881790537) they seem fine, the failure is now not being able to find the version to download

@panticmilos and @IvanZosimov  any new if this is going to be merged? if not I will not bother rebasing it for now

brunoborges:
This branch is large, stale, and conflicted. The feature gap may still exist (`java-version-file` parsing for `pom.xml`/`build.gradle`). Recommendation: decide whether to pursue with a new focused PR.

brunoborges:
Closing as outdated in favor of continuing this work in #1034.

PR #499, opened 2023-06-10
feat: implement cache-dependency-path option to control caching dependency

itchyny:
**Description:**
This PR implements a new option `cache-dependency-path` to control caching dependency.
This option originates from [setup-node](https://github.com/actions/setup-node), and is useful for monorepos.

**Related issue:**

**Check list:**


itchyny:
Hmm, I was worried about the complexity of listing dependency files, but I noticed that [`ruby/setup-ruby`](https://github.com/ruby/setup-ruby) supports `working-directory` option, and that would resolve my use case. There's no actual use case to cache across multiple sub-directories. And the implementation is much simple.
https://github.com/ruby/setup-ruby/blob/v1.151.0/index.js#L45
https://github.com/actions/setup-node/issues/706

nmamlm (approved):

itchyny:
Any chance to get this merged? Despite my previous comment, I think this is the actions/* way since this option is supported not only by setup-node, but setup-go and setup-python.

IvanZosimov:
Hi, @itchyny 👋 Thank you for the contribution! We will take a look and get back to you with our decision regarding it.

ZimvoMatwa (approved):

itchyny:
Oops, I thought the CI failure is something flaky, but it's just about code formatting. I'll fix this soon.

src/cache.ts
```diff
@@ -83,31 +83,34 @@ function findPackageManager(id: string): PackageManager {
...
+  packageManager: PackageManager,
+  cacheDependencyPath?: string
+) {
+  const pattern = cacheDependencyPath
+    ? cacheDependencyPath.trim().split('\n')
+    : packageManager.pattern;
+  const fileHash = await glob.hashFiles(pattern.join('\n'));
+  if (!fileHash) {
+    throw new Error(
+      `No file in ${process.cwd()} matched to [${pattern}], make sure you have checked out the target repository`
```
IvanZosimov, line 98: Let's convert this to warning, as this logic changes the way how action worked previously, by that I mean this comment:
`If there is no file matched to {@link PackageManager.path}, the generated key ends with a dash (-).`

IvanZosimov (changes requested):

IvanZosimov:
Hi @itchyny, I left some comments to your PR, please check them out. Also, I think it'd be great to update the list of inputs in the `action.yml` file and add some e2e tests for the new functionality. Examples of e2e tests you can find in the `./.github/workflows/e2e-cache.yml` file.

src/cache.ts
itchyny: Previously, the error was thrown [here](https://github.com/actions/setup-java/pull/499/files#diff-d027d3b9853e6bedddf63bd1d8fa67c69b4c1ff2b947685108f315e4e831cdbdL104). So I think the action works as same as before.

itchyny:
How can I reliably verify that cache restoration fails on a project after another job of caching different project? The cache seems to be shared across different runs.

IvanZosimov (approved):

dsame (approved):

nikolai-laevskii (approved):

src/cache.ts
```diff
@@ -83,31 +83,34 @@ function findPackageManager(id: string): PackageManager {
...
+    );
+  }
```
IvanZosimov, line 100: May I ask you to remove this logic from there? I think it's better to keep throwing logic from the previous version.

src/cache.ts
```diff
@@ -83,31 +83,34 @@ function findPackageManager(id: string): PackageManager {
...
+  const primaryKey = await computeCacheKey(packageManager, cacheDependencyPath);
   core.debug(`primary key is ${primaryKey}`);
   core.saveState(STATE_CACHE_PRIMARY_KEY, primaryKey);
-  if (primaryKey.endsWith('-')) {
-    throw new Error(
-      `No file in ${process.cwd()} matched to [${
-        packageManager.pattern
-      }], make sure you have checked out the target repository`
-    );
-  }
```
IvanZosimov, line 110: We should keep this, as the post-step of the action, namely save() function depends on the state `STATE_CACHE_PRIMARY_KEY`. If you throw error inside the computeCacheKey() function, this state will not be set.

IvanZosimov (changes requested):

.github/workflows/e2e-cache-dependency-path.yml
```diff
@@ -0,0 +1,93 @@
...
+jobs:
+  gradle1-save:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-latest]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
MaksimZhukov, line 27: Please update checkout version
```suggestion
        uses: actions/checkout@v4
```

itchyny: Would you update the action version along with the other workflows? Thanks.

src/cache.ts
itchyny: I'd like to know why that's necessary before reverting. When there is no gradle (or maven, sbt) file to compute the cache key, we should not save anything on post process, not just restore something. So in that situation, and if `save()` is called, we should make it quickly return, now it does with `Error retrieving key from state.`. But before that, `save()` is never called in this situation; only called on successful job checked in `saveCache()`. Why do you think invalid primary key state should be saved?

IvanZosimov: I agree with you, post-step has `isJobStatusSuccess()` called inside that means that saving state is not required.

IvanZosimov (approved):

MaksimZhukov (approved):

marko-zivic-93 (approved):

dmitry-shibanov (approved):

dusan-trickovic (approved):

PR #500, opened 2023-06-10
This PR adds the instruction to download custom distribution JDK and install

ragsmpl:
**Description:**
Describe your changes.

**Related issue:**

**Check list:**

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -168,13 +168,14 @@ steps:
 - run: java -cp java HelloWorldApp
 ```
 
-If your use-case reuires a custom distribution or a version that is not provided by setup-java and always install the latest version during runtime, then you can use below code to auto download the latest JDK, determine the semver needed for the setup-java and setup-java will take care of the installation and caching on the VM:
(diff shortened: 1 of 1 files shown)


dmitry-shibanov (approved):

marko-zivic-93 (approved):

docs/advanced-usage.md
```diff
@@ -168,6 +168,31 @@ steps:
 - run: java -cp java HelloWorldApp
 ```
 
+If your use-case reuires a custom distribution (in the example alpine-linux is used) or a version that is not provided by setup-java and always install the latest version during runtime, then you can use below code to auto download the latest JDK, determine the semver needed for the setup-java and setup-java will take care of the installation and caching on the VM:
```
IvanZosimov, line 171:
```suggestion
If your use-case requires a custom distribution (in the example, alpine-linux is used) or a version that is not provided by setup-java and you want to always install the latest version during runtime, then you can use the following code to auto-download the latest JDK, determine the semver needed for setup-java, and setup-java will take care of the installation and caching on the VM:
```
I'd suggest to update description like that: 🔼

ragsmpl: Thank you. Updated.

PR #503, opened 2023-06-28
Fix XML namespaces wrongly using https

gnodet:
**Description:**
This PR fix the XML namespaces used when creating the maven toochains.xml file.
The namespaces are identifiers and do not start with 'https', but 'http'.
See https://www.w3.org/TR/xmlschema-1/#Instance_Document_Constructions and 
https://maven.apache.org/xsd/toolchains-1.1.0.xsd.
This is problematic and will break when Maven 4 uses a more conformant XML parser.

**Related issue:**

**Check list:**

### DIFF
--- a/__tests__/toolchains.test.ts
+++ b/__tests__/toolchains.test.ts
@@ -77,9 +77,9 @@ describe('toolchains tests', () => {
     };
 
     const result = `<?xml version="1.0"?>
-<toolchains xmlns="https://maven.apache.org/TOOLCHAINS/1.1.0"
-  xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="https://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd">
+<toolchains xmlns="http://maven.apache.org/TOOLCHAINS/1.1.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd">
   <toolchain>
     <type>jdk</type>
     <provides>
@@ -248,9 +248,9 @@ describe('toolchains tests', () => {
     };
 
     const expectedToolchains = `<?xml version="1.0"?>
-<toolchains xmlns="https://maven.apache.org/TOOLCHAINS/1.1.0"
-  xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="https://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd">
+<toolchains xmlns="http://maven.apache.org/TOOLCHAINS/1.1.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd">
   <toolchain>
     <type>jdk</type>
     <provides>
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -103848,9 +103848,9 @@ function generateToolchainDefinition(original, version, vendor, id, jdkHome) {
     else
         xmlObj = xmlbuilder2_1.create({
             toolchains: {
-                '@xmlns': 'https://maven.apache.org/TOOLCHAINS/1.1.0',
-                '@xmlns:xsi': 'https://www.w3.org/2001/XMLSchema-instance',
-                '@xsi:schemaLocation': 'https://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd',
+                '@xmlns': 'http://maven.apache.org/TOOLCHAINS/1.1.0',
+                '@xmlns:xsi': 'http://www.w3.org/2001/XMLSchema-instance',
+                '@xsi:schemaLocation': 'http://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd',
                 toolchain: [
                     {
                         type: 'jdk',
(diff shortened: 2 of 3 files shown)


gnodet:
The outcome is the following failure: https://github.com/apache/maven/actions/runs/5398872373/jobs/9806118625?pr=1185#step:6:12518
```
Warning: 
Warning:  Some problems were encountered while building the effective toolchains
Warning:  ParseError at [row,col]:[2,1]
Message: Unknown attribute 'schemaLocation' for tag 'toolchains' @ line 2, column 1
Warning: 
```

ZimvoMatwa (approved):

IvanZosimov:
Hi, @gnodet 👋 Are you sure that `https` scheme can't be used and the root cause of your problem is there? If you take a look at the https://www.w3.org/TR/xmlschema-1/#xsi.schemaLocation you may see that it can be of type https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/datatypes.html#anyURI. As the `anyURI` represents a Uniform Resource Identifier Reference, we can find information about URI reference in https://www.ietf.org/rfc/rfc2396.txt. It seems like this standard doesn't limit the number of URI schemas to those listed in the standard, like the `http`. The `https` is just yet another standardised https://datatracker.ietf.org/doc/html/rfc7230#section-2.7.2 URI schema. Correct me if I'm wrong, please.

IvanZosimov:
Just a gentle ping, @gnodet 📟

gnodet:
> Hi, @gnodet 👋 Are you sure that `https` scheme can't be used and the root cause of your problem is there? If you take a look at the https://www.w3.org/TR/xmlschema-1/#xsi.schemaLocation you may see that it can be of type https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/datatypes.html#anyURI. As the `anyURI` represents a Uniform Resource Identifier Reference, we can find information about URI reference in https://www.ietf.org/rfc/rfc2396.txt. It seems like this standard doesn't limit the number of URI schemas to those listed in the standard, like the `http`. The `https` is just yet another standardised https://datatracker.ietf.org/doc/html/rfc7230#section-2.7.2 URI schema. Correct me if I'm wrong, please.

Any valid URI could be used to define _a_ schema URI, you're absolutely right.  However, in order to be a valid maven toolchains xml file, the one _defined by maven_ MUST be used, else the xml simply refers to a different schema namespace.  The same is true for the xsi namespace.  So the namespace identifiers must match those defined by the schemas and if you look at https://www.w3.org/TR/xmlschema-1/#Instance_Document_Constructions and
https://maven.apache.org/xsd/toolchains-1.1.0.xsd, the namespaces are identified with the `http` protocol in their URI.  That's unrelated to the actual schema location  (inside the `xsi:schemaLocation`) which is a URL that just needs to point to the actual xml schema, and thus could use http or https.

Please try to validate the following file with an online xml validator such as https://www.freeformatter.com/xml-validator-xsd.html, you should end up with an error `Cvc-elt.1.a: Cannot Find The Declaration Of Element 'toolchains'., Line '3', Column '116'.` because the `toolchains` element is not part of the `https://maven.apache.org/TOOLCHAINS/1.1.0` namespace, but the `http://maven.apache.org/TOOLCHAINS/1.1.0`
```
<?xml version="1.0"?>
<toolchains xmlns="https://maven.apache.org/TOOLCHAINS/1.1.0"
  xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="https://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd">
  <toolchain>
    <type>jdk</type>
    <provides>
      <version>17</version>
      <vendor>Eclipse Temurin</vendor>
      <id>temurin_17</id>
    </provides>
    <configuration>
      <jdkHome>/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.1-12/x64</jdkHome>
    </configuration>
  </toolchain>
</toolchains>
```

Using the fixed xml below pass the validation successfully:
```
<?xml version="1.0"?>
<toolchains xmlns="http://maven.apache.org/TOOLCHAINS/1.1.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd">
  <toolchain>
    <type>jdk</type>
    <provides>
      <version>17</version>
      <vendor>Eclipse Temurin</vendor>
      <id>temurin_17</id>
    </provides>
    <configuration>
      <jdkHome>/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.1-12/x64</jdkHome>
    </configuration>
  </toolchain>
</toolchains>
```

gnodet:
Fwiw, maven 3 does not validate the XML when they are loaded.   That's why this issue has not caused any real issue at this point. However maven 4 will reject any unknown attribute, hence this PR to fix the problem before it actually hits people.

gnodet:
To be more precise, the error I mentioned above from the validator is related to the fact that the `xsi:schemaLocation` is simply ignored because `xsi`  points to the wrong namespace.

If we fix the `xsi` namespace with the below file, a bunch of different and more precise errors will be printed: `TargetNamespace.1: Expecting Namespace 'https://maven.apache.org/TOOLCHAINS/1.1.0', But The Target Namespace Of The Schema Document Is 'http://maven.apache.org/TOOLCHAINS/1.1.0'., Line '24', Column '195'.` because of the mismatch between the defined namespace and the one that is actually used.

```
<?xml version="1.0"?>
<toolchains xmlns="https://maven.apache.org/TOOLCHAINS/1.1.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="https://maven.apache.org/TOOLCHAINS/1.1.0 https://maven.apache.org/xsd/toolchains-1.1.0.xsd">
  <toolchain>
    <type>jdk</type>
    <provides>
      <version>17</version>
      <vendor>Eclipse Temurin</vendor>
      <id>temurin_17</id>
    </provides>
    <configuration>
      <jdkHome>/opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.1-12/x64</jdkHome>
    </configuration>
  </toolchain>
</toolchains>
```

IvanZosimov:
Huge thanks for the clarification, @gnodet 👋

IvanZosimov (approved):

marko-zivic-93 (approved):

dusan-trickovic (approved):

MaksimZhukov (approved):

dmitry-shibanov (approved):

PR #507, opened 2023-07-05
Update Oracle JDK download URL calculation

nikolai-laevskii:
**Description:**
Updates how Oracle JDK download URL is determined.
Previously it would pick either "archive" or "latest" download links depending on whether user specified just major version (it would select "latest") or major.minor.(patch) as well (it would select "archive"). It doesn't seem like a correct logic because it's just the version format, it shouldn't determine whether it should be taken from the archive or from latest. Moreover, links for latest versions are invalidated often which causes tests to fail and may potentially break workflows of some people.

I propose just picking every version from the "archive" as even latest versions added there immediately and archive links don't seem to change that often. 

**Related issue:**
No particular issue

**Check list:**


fniephaus:
>  It doesn't seem like a correct logic because it's just the version format, it shouldn't determine whether it should be taken from the archive or from latest.

The reason for this is to allow users to set `java-version` to a major number. With this, users must always use the full version number when using Oracle JDK.

If we are moving forward with this PR, then you may also want to update [this code snippet](
https://github.com/actions/setup-java/blob/1f2faad7e0bc7efd1f6da4ca0a0f2b23b03a3a7d/docs/advanced-usage.md#Oracle).

dsame (approved):

marko-zivic-93 (approved):

nikolai-laevskii:
@fniephaus 
> > It doesn't seem like a correct logic because it's just the version format, it shouldn't determine whether it should be taken from the archive or from latest.
> 
> The reason for this is to allow users to set `java-version` to a major number. With this, users must always use the full version number when using Oracle JDK.
> 
> If we are moving forward with this PR, then you may also want to update [this code snippet](https://github.com/actions/setup-java/blob/1f2faad7e0bc7efd1f6da4ca0a0f2b23b03a3a7d/docs/advanced-usage.md#Oracle).

Thank you for noting this! Could you please elaborate on this a bit further? As of now archive seems to allow downloading with only a major version specified.

fniephaus:
> Thank you for noting this! Could you please elaborate on this a bit further? As of now archive seems to allow downloading with only a major version specified.

You are right, I wasn't aware that these links also work:
- https://download.oracle.com/java/17/archive/jdk-17_linux-x64_bin.tar.gz
- https://download.oracle.com/java/20/archive/jdk-20_linux-x64_bin.tar.gz

Thanks for the PR!

fniephaus (approved):
LGTM

PR #522, opened 2023-08-09
Make the Adoptopenjdk package type look at the Temurin repo first for latest assets

johnoliver:
We are looking to move people off of the old deprecated Adoptopenjdk API to the Temurin API for Hotspot builds.

For the vast majority of binaries, the Temurin API will contain the one you want, the outlier cases are:
- User is using an older (<16) non-lts release
- User is requesting a very specific version from before Temurin

In both cases it means that they are using very old builds.

In order to move to the new API, this PR is proposing that requests to AdoptDistribution, are requested against TemurinDistribution first, if they cannot be satisfied by Temurin, then continue the existing unchanged AdoptOpenJDK logic. For the VAST majority of requests, they will be satisfied by Temurin, for the few remaining outliers described above, they will make a second request to the old AdoptOpenJDK API.

I am submitting this as a draft to ask if this approach is acceptable in principle.

### DIFF
--- a/__tests__/distributors/adopt-installer.test.ts
+++ b/__tests__/distributors/adopt-installer.test.ts
@@ -4,6 +4,7 @@ import {
   AdoptDistribution,
   AdoptImplementation
 } from '../../src/distributions/adopt/installer';
+import {TemurinDistribution} from '../../src/distributions/temurin/installer';
 import {JavaInstallerOptions} from '../../src/distributions/base-models';
 
 import os from 'os';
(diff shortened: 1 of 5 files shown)


gdams:
I'm wondering whether we could add an annotation the pipeline to alert users to the fact that they should switch to temurin as their distro if the version they have requested is matched in temurin.

IvanZosimov:
Hi, @johnoliver and @gdams 👋 Actually, we have a piece of documentation about this problem: https://github.com/actions/setup-java#supported-distributions. Frankly speaking, we'd better wait for the official announcement of the full deprecation of the AdoptOpenJDK; after that, we could make an announcement for our customers. Given that many customers are aware of the deprecation but continue to use the outdated versions, the warning option also doesn't seem to be the best as it could create a lot of noise.

gdams:
> Hi, @johnoliver and @gdams 👋 Actually, we have a piece of documentation about this problem: https://github.com/actions/setup-java#supported-distributions. Frankly speaking, we'd better wait for the official announcement of the full deprecation of the AdoptOpenJDK; after that, we could make an announcement for our customers. Given that many customers are aware of the deprecation but continue to use the outdated versions, the warning option also doesn't seem to be the best as it could create a lot of noise.

@IvanZosimov If you're looking for an official deprecation you can consider this blog post exactly that: https://blog.adoptopenjdk.net/2021/08/goodbye-adoptopenjdk-hello-adoptium/. The challenge here is that there are several Java versions that we released at AdoptOpenJDK that we don't release at Adoptium (e.g JDK 12-15 and 9-10). The idea of this change is that it will preserve functionality for those specific users but get the majority of users onto the newer API. As a side note, we should have a discussion with the dependabot team at some point to work out a plan for updating people's workflows.

gdams:
@IvanZosimov what are your thoughts here? We'd love to find a solution to take some of the requests off our old API

gdams:
@johnoliver looks like you need to run prettier and generate the `dist` files

gdams:
ping @IvanZosimov it would be good to make some progress here

IvanZosimov:
Hi, @gdams and @johnoliver 👋 I apologize for taking so long to respond, we are overwhelmed at the moment! We discussed the idea of PR, and it seems acceptable to us. I think you can convert this PR to the real one, and we will take a look at it as we have time.

johnoliver:
I have just noticed that the tests need fixing, will get on updating this

johnoliver:
Tests are fixed

gdams:
@IvanZosimov this is good for review now

Calzadas (changes requested):
ldkdA

gdams:
@aparnajyothi-y any update here please?

gdams (approved):

PR #302, opened 2022-03-24
Add SBT in cache managers.

fmeriaux:
**Description:**
Describe your changes.

Add SBT cache manager.

**Related issue:**
Add link to the related issue.

https://github.com/actions/setup-java/issues/266#issuecomment-1077368457

**Check list:**
- [x] Mark if documentation changes are required.
- [ ] Mark if tests were added or updated to cover the changes.

PR #564, opened 2023-12-06
Avoid `sbt` cache being invalidated for a library that is only incrementing its own version

rtyley:
**Description:**
While working on [`guardian/gha-scala-library-release-workflow`](https://github.com/guardian/gha-scala-library-release-workflow) I noticed that no matter how many times I ran the workflow, `actions/setup-java` would always report [`sbt cache is not found`](https://github.com/guardian/play-secret-rotation/actions/runs/7117908339/job/19381358729), even if there had been no substantial change in the project - simply that the version number in `version.sbt` (the file used by [`sbt/sbt-release`](https://github.com/sbt/sbt-release)) had been incremented (as in https://github.com/guardian/play-secret-rotation/commit/b2152325baf80f57e7c510fe03ccda5596c139d6).

This meant that turning on [`cache: sbt`](https://github.com/actions/setup-java#caching-sbt-dependencies) would actually slow the workflow considerably, as it would never benefit from the cache being present, and would always have to save it, which could take 2-3 minutes - even though it can't take advantage of the data it's saving.

As such, it would be great to exclude `version.sbt` files from the cache hash key.

Background: `cache: sbt` was originally introduced with https://github.com/actions/setup-java/pull/302 - cc @fmeriaux 🙇‍♂️ 

**Check list:**

### DIFF
--- a/dist/cleanup/index.js
+++ b/dist/cleanup/index.js
@@ -319,7 +319,8 @@ function createHttpClient() {
     return new http_client_1.HttpClient('actions/cache', [bearerCredentialHandler], getRequestOptions());
 }
 function getCacheVersion(paths, compressionMethod, enableCrossOsArchive = false) {
-    const components = paths;
+    // don't pass changes upstream
+    const components = paths.slice();
     // Add compression method to cache version to restore
     // compressed cache as per compression method
     if (compressionMethod) {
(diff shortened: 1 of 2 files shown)


fmeriaux:
Interesting, perhaps it would be a good idea to have an action parameter that allows you to complete the pattern by default? 

This would give us finer control over the specifics of our projects.

brunoborges:
The sbt cache invalidation concern appears valid, but this branch is stale/outdated. Recommendation: decide whether to pursue with a minimal fresh PR against current `main`.

brunoborges:
Seeing that the SBT project maintains a Setup SBT action, I am closing this stale PR.

https://github.com/sbt/setup-sbt

rtyley:
> Seeing that the SBT project maintains a Setup SBT action, I am closing this stale PR.

Could you re-open this PR? Although the SBT community do maintain a `setup-sbt` GitHub Action, that action depends on `setup-java`, as can be seen in https://github.com/sbt/setup-sbt#usage :

```yaml
steps:
- uses: actions/checkout@v6
- name: Setup JDK
  uses: actions/setup-java@v5
  with:
    distribution: temurin
    java-version: 17
    cache: sbt
- uses: sbt/setup-sbt@v1
- name: Build and test
  shell: bash
  run: sbt -v +test
```

...the recommended usage of `setup-sbt` depends on using `actions/setup-java` and it's `cache: sbt` parameter, so fixing the caching problems of `setup-java` is still important.

Issue #600, opened 2024-02-25
Using GPG on V4, the gpg importer will always take 3 minutes


GedMarc:
**Description:**
When configuring and using setup-java on v4 with gpg parameters, the step that imports the gpg key will always take 3 minutes

**Task version:**
V4

**Platform:**
- [ X] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [ X] Hosted
- [ ] Self-hosted

**Repro steps:**  
Create a working GPG setup using setup-java on v3. Update the version to v4. The task will now take 3 minutes. 

Working 
```
       - name: Set up Java
        uses: actions/setup-java@v3
        with:
          java-version: ${{ matrix.java-version }}
          distribution: 'oracle'
          #server-id: github_token
          #settings-path: ${{ github.workspace }}
          cache: 'maven'
          gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
          gpg-passphrase: GPG_PASSPHRASE
```
Affected
```
      - name: Set up Java
        uses: actions/setup-java@v4
        with:
          java-version: ${{ matrix.java-version }}
          distribution: 'oracle'
          #server-id: github_token
          #settings-path: ${{ github.workspace }}
          cache: 'maven'
          gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
          gpg-passphrase: secrets.GPG_PASSPHRASE
```
**Expected behavior:**
The same as in V3, when the task is complete the job must continue immediately. 

**Actual behavior:**
The step waits for 3 minutes on any workflow for the task to complete

HarithaVattikuti:
Hello @GedMarc 
Thank you for creating this issue. We will investigate it and get back to you as soon as we have some feedback.

priyagupta108:
Hello @GedMarc  👋 ,
We've released a new version of `setup-java` ([v4.2.0](https://github.com/actions/setup-java/releases/tag/v4.2.0)), and the v4 tag has been updated accordingly. This update includes a version bump for `actions/http-client` to `2.2.1`. We anticipate that this change might address the delay issue with the GPG importer.
Could you please confirm if it resolves your issue?

priyagupta108:
Hey @GedMarc, Just checking in to see if your issue has been addressed with the recent setup-java update ([v4.2.0](https://github.com/actions/setup-java/releases/tag/v4.2.0)). Could you please update us on this? Thank you in advance for your update!

GedMarc:
[...]

PR #608, opened 2024-03-13
feat: Update recommended configuration for GPG signing

wetneb:
This attempts to document the new recommended configuration to sign artifacts with the maven-gpg-plugin as part of the deploy process.

It imitates this PR from the maintainer of the maven-gpg-plugin: https://github.com/xerial/sqlite-jdbc/pull/1082/files

Notes that this requires the maven-gpg-plugin version 3.2.0 or above, not sure if this is worth adding to the documentation as I expect this guide will mostly be followed by people setting up a new project (hopefully using the latest version of the plugin by default).

@cstamas I hope I got it right, feel free to suggest any improvements

**Related issue:**
might be related to #600?

see also https://issues.apache.org/jira/browse/MGPG-90?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=17825880

**Check list:**


cstamas (approved):
LGTM, but this would work with 3.2.0 m-gpg-p only of course.

wetneb:
With the release of the 3.2.1 version of the maven-gpg-plugin, this documentation update is less important as 3.2.1 will continue to work like the previous versions.

However, on the long term, this would likely still be a better choice than the current set up, as it removes the dependency on an external GPG agent. Arguably setup-java shouldn't have anything to do with setting up a GPG environment as it's unrelated to Java.

One downside of this method is the additional `-Dgpg.signer=bc` that needs to be added to `mvn package`. Perhaps that's something that could be removed in the future, if the maven-gpg-plugin is able to detect by itself that it is running in an environment where GPG hasn't been set up at all.

brunoborges:
The recommendation is useful, but current docs intentionally cover the setup-java-managed GPG path. Recommendation: decide whether to document the BC signer flow as an additional/alternative path in a fresh docs PR.

brunoborges (approved):

brunoborges (approved):

Issue #700, opened 2024-10-22
Could not find GraalVM for SemVer 17


violetagg:
**Description:**
The issue below started to appear on 16.10.2024 without any changes in the project

```
Trying to resolve the latest version from remote
  Error: Could not find GraalVM for SemVer 17
```

**Task version:**
v4

**Platform:**
- [X] Ubuntu
- [X] macOS
- [X] Windows

**Runner type:**
- [X] Hosted
- [ ] Self-hosted

**Repro steps:**  
https://github.com/reactor/reactor-netty/actions/runs/11357786129

```
      - name: Set up GraalVM 17
        uses: actions/setup-java@b36c23c0d998641eff861008f374ee103c25ac73
        with:
            distribution: 'graalvm'
            java-version: '17'
```

**Expected behavior:**
GraalVM to be installed

**Actual behavior:**
GraalVM is not installed

rubarcelk:
Hey, any updated about this issue?

priyagupta108:
Hello @violetagg,
The action uses the [latest URLs](https://github.com/actions/setup-java/blob/83a06ff9d9aa70f76a8d73278e646c20b2bf1ae5/src/distributions/graalvm/installer.ts#L85) to resolve `java-version: '17'` for Oracle GraalVM. Unfortunately, these URLs are no longer available. As a result, specifying `java-version: '17'` no longer works. However, you can specify `java-version: '17.0.12'`, which will pull from the [archive](https://github.com/actions/setup-java/blob/83a06ff9d9aa70f76a8d73278e646c20b2bf1ae5/src/distributions/graalvm/installer.ts#L82) and function correctly.

Please feel free to reach out to us in case of any concerns.

violetagg:
@priyagupta108 Thanks for the suggestion. It is working now https://github.com/reactor/reactor-netty/actions/runs/11493523710
We can close this.

Marcono1234:
The underlying issue seems to be that licensing for GraalVM for JDK 17 changed in September 2024, and newer JDK 17 versions are not available anymore under the "GraalVM Free Terms and Conditions (GFTC)". See https://github.com/actions/setup-java/pull/501#discussion_r1803652563 and related discussion in https://github.com/graalvm/setup-graalvm/issues/107.

[...]

PR #704, opened 2024-11-07
Add Documentation to Recommend Using GraalVM JDK 17 Version to 17.0.12 to Align with GFTC License Terms

aparnajyothi-y:
**Description:**
This PR will update to recommend using graalvm 17 version to comply with the GraalVM free terms and conditions license(GFTC).

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,9 @@ Currently, the following distributions are supported:
 
 **NOTE:** For Azul Zulu OpenJDK architectures x64 and arm64 are mapped to x86 / arm with proper hw_bitness.
 
+**NOTE:** To comply with the GraalVM Free Terms and Conditions (GFTC) license, it is recommended to use GraalVM JDK 17 version
+ ' 17.0.12 ', as this is the only version of GraalVM JDK 17 that is available under the GFTC license.
+
 ### Caching packages dependencies
 The action has a built-in functionality for caching and restoring dependencies. It uses [toolkit/cache](https://github.com/actions/toolkit/tree/main/packages/cache) under hood for caching dependencies but requires less configuration settings. Supported package managers are gradle, maven and sbt. The format of the used cache key is `setup-java-${{ platform }}-${{ packageManager }}-${{ fileHash }}`, where the hash is based on the following files:
 
--- a/README.md
+++ b/README.md
@@ -117,8 +117,7 @@ Currently, the following distributions are supported:
 
 **NOTE:** For Azul Zulu OpenJDK architectures x64 and arm64 are mapped to x86 / arm with proper hw_bitness.
 
-**NOTE:** To comply with the GraalVM Free Terms and Conditions (GFTC) license, it is recommended to use GraalVM JDK 17 version
- ' 17.0.12 ', as this is the only version of GraalVM JDK 17 that is available under the GFTC license.
+**NOTE:** To comply with the GraalVM Free Terms and Conditions (GFTC) license, it is recommended to use GraalVM JDK 17 version 17.0.12, as this is the only version of GraalVM JDK 17 available under the GFTC license. Additionally, it is encouraged to consider upgrading to GraalVM JDK 21, which offers the latest features and improvements.
 
 ### Caching packages dependencies
 The action has a built-in functionality for caching and restoring dependencies. It uses [toolkit/cache](https://github.com/actions/toolkit/tree/main/packages/cache) under hood for caching dependencies but requires less configuration settings. Supported package managers are gradle, maven and sbt. The format of the used cache key is `setup-java-${{ platform }}-${{ packageManager }}-${{ fileHash }}`, where the hash is based on the following files:


fniephaus:
Hi @aparnajyothi-y, thanks for the PR. I work on the GraalVM at Oracle and contributed GraalVM support.

Instead of adding a note like this, we are using a different approach in [`setup-graalvm`](https://github.com/graalvm/setup-graalvm): we let [`java-version: '17'` default to `17.0.12` and emit a job warning](https://github.com/graalvm/setup-graalvm/blob/557ffcf459751b4d92319ee255bf3bec9b73964c/src/graalvm.ts#L36-L40).

Do you think it would make sense to do the same in `setup-java`?

aparnajyothi-y:
Hello @fniephaus, Thank you for the suggestion. Based on further investigation, it wouldn’t be feasible to adopt the same approach for Java 17 as in setup-graalvm at this time, because the end of GFTC support for version 17.0.12 has not yet been confirmed. As a result, we cannot be certain whether this will be the final update of Java 17 with GFTC license support, or if future versions will follow the same model.

Given this uncertainty, we’ve decided to update the documentation instead of implementing the same defaulting approach as in setup-graalvm.

We appreciate your understanding!

noob-code55 (approved):
- name: Setup Java JDK
  uses: actions/setup-java@v4.6.0
  with:
    # The Java version to set up. Takes a whole or semver Java version. See examples of supported syntax in README file
    java-version: # optional
    # The path to the `.java-version` file. See examples of supported syntax in README file
    java-version-file: # optional
    # Java distribution. See the list of supported distributions in README file
    distribution: 
    # The package type (jdk, jre, jdk+fx, jre+fx)
    java-package: # optional, default is jdk
    # The architecture of the package (defaults to the action runner's architecture)
    architecture: # optional
    # Path to where the compressed JDK is located
    jdkFile: # optional
    # Set this option if you want the action to check for the latest available version that satisfies the version spec
    check-latest: # optional
    # ID of the distributionManagement repository in the pom.xml file. Default is `github`
    server-id: # optional, default is github
    # Environment variable name for the username for authentication to the Apache Maven repository. Default is $GITHUB_ACTOR
    server-username: # optional, default is GITHUB_ACTOR
    # Environment variable name for password or token for authentication to the Apache Maven repository. Default is $GITHUB_TOKEN
    server-password: # optional, default is GITHUB_TOKEN
    # Path to where the settings.xml file will be written. Default is ~/.m2.
    settings-path: # optional
    # Overwrite the settings.xml file if it exists. Default is "true".
    overwrite-settings: # optional, default is true
    # GPG private key to import. Default is empty string.
    gpg-private-key: # optional
    # Environment variable name for the GPG private key passphrase. Default is $GPG_PASSPHRASE.
    gpg-passphrase: # optional
    # Name of the build platform to cache dependencies. It can be "maven", "gradle" or "sbt".
    cache: # optional
    # The path to a dependency file: pom.xml, build.gradle, build.sbt, etc. This option can be used with the `cache` option. If this option is omitted, the action searches for the dependency file in the entire repository. This option supports wildcards and a list of file names for caching multiple dependencies.
    cache-dependency-path: # optional
    # Workaround to pass job status to post job step. This variable is not intended for manual setting
    job-status: # optional, default is ${{ job.status }}
    # The token used to authenticate when fetching version manifests hosted on github.com, such as for the Microsoft Build of OpenJDK. When running this action on github.com, the default value is sufficient. When running on GHES, you can pass a personal access token for github.com if you are experiencing rate limiting.
    token: # optional, default is ${{ github.server_url == 'https://github.com' && github.token || '' }}
    # Name of Maven Toolchain ID if the default name of "${distribution}_${java-version}" is not wanted. See examples of supported syntax in Advanced Usage file
    mvn-toolchain-id: # optional
    # Name of Maven Toolchain Vendor if the default name of "${distribution}" is not wanted. See examples of supported syntax in Advanced Usage file
    mvn-toolchain-vendor: # optional

aparnajyothi-y:
Hello @fniephaus, a gentle reminder to check if you have any further questions or need additional clarifications on the above.

fniephaus:
Thanks for the ping, @aparnajyothi-y!

The [blog post that introduced GFTC](https://blogs.oracle.com/java/post/graalvm-free-license) said the following about timelines:

"For designated Long Term Support releases (GraalVM for JDK 17), Oracle will provide these free GFTC releases until one full year after the subsequent LTS release".

According to [the release calendar](https://docs.oracle.com/en/graalvm/enterprise/22/docs/release-calendar/), Oracle GraalVM for JDK 17.0.12 is the last release under GFTC. The artifacts are still available and their license are not changed retroactively. In fact, the [script-friendly URLs](https://www.oracle.com/java/technologies/jdk-script-friendly-urls/) used by `setup-java` can only retrieve GFTC-licensed releases of Oracle GraalVM.

The [end of permissive license was also announced here](https://blogs.oracle.com/java/post/jdk-17-approaches-endofpermissive-license).

So, I'd still recommend to let 'java-version: 17' fall back to 17.0.12 while displaying a warning that notifies users of their options (upgrading to JDK 21 is recommended).

What do you think?

Marcono1234:
Related issue: #700

mahabaleshwars (approved):

aparnajyothi-y:
Hello @fniephaus,

Thank you for highlighting that Oracle GraalVM for JDK 17.0.12 is the final GFTC release from the blog. It's reassuring to know that the artifacts and licenses remain unchanged, and that the setup-java URLs will continue to point to GFTC-licensed releases.

I agree with your suggestion to have java-version: 17 fall back to 17.0.12, while displaying a warning to users and encouraging them to consider upgrading to JDK 21. As for making any direct changes to 17.0.12, we'll likely monitor feedback from the community moving forward to determine if any adjustments are necessary. This approach will help us make data-driven decisions based on the needs of the majority of users.

We encourage your future feedback as we continue to improve setup-java.

priya-kinthali (approved):

HarithaVattikuti (approved):

PR #750, opened 2025-02-11
Issue725 read me supported version syntax

### LINKED ISSUE #725: Support all Java versions
**Description:**

The README mentions only the Java versions "8, 11, 16, 17, 21". [1]

Can you please explain the rationale for how it is decided to support a certain version or not? 

I would additionally request adding support for all Java versions (at least going forward) because "The Tip & Tail Model of Library Development" [2] recommended by OpenJDK would require popular libraries to be able to build and release against the latest versions (as well as using older versions to verify compatibility of older releases).

[1] https://github.com/actions/setup-java/blob/main/README.md#supported-version-syntax
[2] https://openjdk.org/jeps/14 

**Justification:**

Supporting more Java version will allow even more maintainers and vendors to be able to use this action for their builds.

skateball:
**Description:**
some small changes on README.md to enhance readabilty

**Related issue:**
fixes https://github.com/actions/setup-java/issues/725 

**Check list:**
- [no] Mark if documentation changes are required.
- [no] Mark if tests were added or updated to cover the changes.


skateball:
Hi @fniephaus 
did a small change in a markdown file and afterwards the test  https://github.com/actions/setup-java/actions/runs/13272353142/job/37054560743?pr=750  runs in an error because downloading  https://download.oracle.com/graalvm/18/latest/graalvm-jdk-18_macos-x64_bin.tar.gz  runs in a 404

We're sorry, the page you requested was not found.
We have recorded this error (404) to help us fix the problem.
You may wish to try again using one of the tools below. 

Can you please check?
Regards Christian

fniephaus:
Hi @skateball,

I'm not sure why this is failing but this seems to be a transient issue. GraalVM for JDK 18 indeed does not exist and the test is supposed to check for exactly that. Instead of an exception being thrown, however, I see this in the log:

> Received promise resolved instead of rejected

I'd assume this is unrelated to your markdown changes.

brunoborges:
This PR seems to be staled and the proposed change is already covered in the current version of the README.

@skateball can you please confirm?

PR #813, opened 2025-04-29
Extended settings.xml - Profiles with Repo

Scyjin:
Building Maven Project with dependencies of another Repository then Maven-Central.

This could be extended to support more than one repo.
Also the profile could be cut down to one overall-profile, i think.

I'm more the Backend Guy, hope its ok. :)

**Description:**
Extended settings.xml generation to include another Repository. 
This should provide nessecary Information while building with Maven.
So your build could refer to something like Artifacts in a private github repository.

**Related issue:**

**Check list:**


brunoborges:
Thank you for this contribution, @Scyjin — the idea of letting `setup-java` configure dependency-resolution repositories in the generated `settings.xml` is genuinely useful, and your prototype helped shape it.

I'm going to close this PR because the branch is now quite stale (April 2025) and conflicts with the recent refactor of `generate()` in `src/auth.ts` (see #1037, which restructures server generation around a structured definition). Rather than rebasing the original approach, I've captured the idea — generalized to support multiple repositories, Central inclusion/ordering, and per-repo snapshot control — in a fresh tracking issue:

➡️ **#1054**

That issue also notes how this composes with the credentials work in #1037 (the `<repositories>` half here pairs with the `<servers>` half there). A reworked implementation on top of current `main` — with tests, docs, and a structured input — would be very welcome there. Thanks again!

PR #819, opened 2025-05-05
Add caching for the Mill build tool

clintval:
**Description:**

Similar to Gradle, Maven, and sbt, I am hoping to see [Mill](https://mill-build.org/mill/index.html) cache support.

Mill is becoming a popular JVM build tool so I hope this is a welcome addition.

I have not contributed to this project before so please let me know if I missed anything!

Speaking of, how do I run automated tests on this branch?

**Check list:**


.github/workflows/e2e-cache.yml
```diff
@@ -207,3 +207,93 @@ jobs:
...
+        uses: actions/checkout@v4
+      - name: Run setup-java with the cache for mill
+        uses: ./
+        id: setup-java
+        with:
+          distribution: 'adopt'
+          java-version: '11'
+          cache: mill
+      - name: Create files to cache
+        run: ./mill --disable-ticker _.compile
```
lihaoyi, line 231: `_.compile` is probably wrong (should be `__.compile`), but we should make this warmup command somethign the user passes in. Some users may want to cache jars, assemblies, or other artifacts beyond just plain compilation

clintval: Thanks for reviewing @lihaoyi!

`_.compile` should be OK here since this is just an integration for a tiny test Scala project:

https://github.com/actions/setup-java/blob/18d114c65a677ef42cfaec1e63c0c7dd1aac9baa/__tests__/cache/mill/build.sc#L1-L12

I can change it to `__.compile` for safety-sake anyway, though.

And this isn't where the magic happens since these are just tests for this project.

The default caching behavior is set to:

https://github.com/actions/setup-java/blob/18d114c65a677ef42cfaec1e63c0c7dd1aac9baa/src/cache.ts#L64-L78

So that anything under `~/.cache/mill` is cached.

The cached is rebuilt if anything in these patterns _**changes**_:

```yaml
pattern: [
  // https://github.com/coursier/cache-action/blob/4e2615869d13561d626ed48655e1a39e5b192b3c/README.md?plain=1#L28-L38
  '**/build.sc',
  '**/*.sc',
  '**/mill',
  '**/.mill-version',
  // https://github.com/com-lihaoyi/mill/blob/5b88d1e268e6264e44589c5ac82c0fdbd680fd63/mill#L6-L11
  '**/.config/mill-version'
]
```

brunoborges:
We will wait a little longer for wider adoption for Mill before considering it.

PR #827, opened 2025-05-12
Added support for configuring multiple maven repositories with separate usernames and passwords

sudiptosarkar:
**Description:**
This PR adds support for configuring multiple maven repositories with separate usernames and passwords for each. Currently the action only supports one repository configuration, which makes it impossible to specify more than one (release and snapshot repositories for example). That would inevitably lead to changing the maven config in the action yaml every time the version changes to/from release/snapshot.

**Related issue:**
#828

**Check list:**


brunoborges:
Linking this PR to the tracked issue: #828 (Need support for multiple maven repositories).

brunoborges:
Maintainer design follow-up: I implemented an alternative in #1037 with a tighter surface area.

Why this design: #827 introduces indexed input tuples (for example server-id-0/server-username-0/server-password-0), which scale poorly and increase maintenance and docs complexity as the number of repositories grows.

What #1037 does differently:
- adds a single structured multiline input: mvn-server-credentials
- one line per server in the format: server-id:USERNAME_ENV:PASSWORD_ENV
- keeps full backward compatibility by preserving the existing single-server inputs as fallback
- includes explicit validation for malformed entries and duplicate server IDs

Net effect: same multi-repository capability with less API complexity and clearer long-term maintainability.

Issue #904, opened 2025-08-30
[DEP0040] DeprecationWarning: The `punycode` module is deprecated


asarkar:
**Description:**
Emits the following warning during set-up and post set-up steps:
```
(node:2563) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

**Task version:**
5

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Repro steps:**  
N/A

**Expected behavior:**
No warnings.

**Actual behavior:**
Warning shown.

aparnajyothi-y:
Hello @asarkar, Thank you for creating this issue and we will look into it :)

aparnajyothi-y:
Hello @asarkar, thanks for reporting this!

We’ve investigated the issue and found the cause: we recently upgraded setup-java’s runtime from Node 20 → Node 24. As per this documentation from [node js](https://nodejs.org/api/punycode.html), starting with Node 24, any code that loads the built-in [punycode](https://nodejs.org/api/punycode.html)
 module triggers a [DEP0040] DeprecationWarning. One of our transitive dependencies, uri-js@4.4.1, imports punycode, which leads to this warning. Even though a userland version of punycode (2.3.1) is installed, Node resolves the core module name first, so the warning still appears.

Workaround:
You can suppress deprecation warnings in your workflow by setting:

env:
  NODE_OPTIONS: --no-deprecation


We are also tracking options to remove or replace the uri-js dependency since it is no longer actively maintained. Until then, the warning can safely be ignored or suppressed as shown above.
please feel free to reach us if any further concerns or clarifications :)

reneleonhardt:
If the abandoned uri-js still is so important, why not switch to a maintained fork?
https://github.com/andreinwald/uri-js-replace

> - Solves "The punycode module is deprecated" warning in Node

PR #917, opened 2025-09-16
Remove punycode deprecation warning by replacing related dependency and upgrading transitive dependencies

aparnajyothi-y:
**Description:**

This change eliminates the Node.js deprecation warning caused by the punycode module.

- Replaced the affected dependency (uri-js) with a maintained alternative.

- Upgraded transitive dependencies to versions.

- Verified that builds and runtime behavior remain stable.

This update keeps the dependency tree aligned with current Node.js standards and removes unnecessary deprecation warnings.

**Related issue:**
[#904](https://github.com/actions/setup-java/issues/904)


tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
-    "target": "es6",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
+    "target": "es2018",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
```
jsoref: It feels like the comment wants to be aligned with its siblings

Issue #937, opened 2025-10-15
SapMachine URL to the releases JSON will change


skateball:
**Description:**
The URL to the Version JSON of the available SapMachine releases will change. Therefore, we want to update it. Please merge soon.
Currently, both links will work. The change will be unnoticed by users.
All platforms/variations affected.
No documentation and no tests required.
I will provide a PR in a few seconds.
Best Regards
Christian@SapMachine-Team at SAP

**Task version:**
Will provide a PR in some seconds

**Platform:**
- [x] Ubuntu
- [x] macOS
- [x] Windows

**Runner type:**
- [x] Hosted
- [x] Self-hosted

**Repro steps:**  
Will provide a PR in some seconds

**Expected behavior:**
Will provide a PR in some seconds

**Actual behavior:**
Will provide a PR in some seconds

mahabaleshwars:
Hi @skateball, closing this as resolved via [#955](https://github.com/actions/setup-java/pull/955), which updates the internal URL used to fetch the SapMachine releases JSON. 

**Summary of what changed:**

Updated the metadata source to the new SapMachine endpoint (the old URL currently still responds, so this is a seamless transition). The modification is non-breaking: both the previous and new endpoints currently function, ensuring uninterrupted version resolution.

PR #965, opened 2025-11-28
Update SapMachine URLs (v4)

RealCLanger:
**Description:**
Updates URLs to SapMachine for v4 release branch

**Related issue:**
#937 

**Check list:**

### DIFF
--- a/dist/cleanup/index.js
+++ b/dist/cleanup/index.js
@@ -19167,7 +19167,7 @@ FormData.prototype.submit = function (params, cb) {
         request.removeListener('error', callback);
         request.removeListener('response', onResponse);
 
-        return cb.call(this, error, responce); // eslint-disable-line no-invalid-this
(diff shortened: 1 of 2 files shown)


RealCLanger:
@mahabaleshwars Is this needed? Are there planned updates for relase v4?

mahabaleshwars:
Hi @RealCLanger, yes, this is needed.

According to issue #937, both the old and new URLs currently work, but this change ensures compatibility when the old URL is eventually deprecated.

Since many users are still on v4 (the latest being v5. 0.0 released in August 2025), we should apply this fix to the releases/v4 branch to prevent disruptions for those users.

RealCLanger:
> Hi @RealCLanger! Thanks for checking in.
> 
> The change is now available in setup-java v4 as of **v4.8.0**. You can use `actions/setup-java@v4` to access the update.

That's great, thank you!

PR #987, opened 2026-03-09
feat: Add cache-write input for read-only cache mode

salmanmkc:
When `cache` is set to maven/gradle/sbt, this action restores and saves the dependency cache. For PR workflows there's no way to get restore-only — a malicious PR can poison the cache and have it picked up by trusted runs on main.

This adds a `cache-write` input (defaults to `true`, backward compatible). When `false`, the cleanup step skips cache saving entirely.

**Usage:**
```yaml
- uses: actions/setup-java@v5
  with:
    distribution: temurin
    java-version: 21
    cache: maven
    cache-write: ${{ github.event_name != 'pull_request' }}
```

**What changed:**
- `action.yml` — new `cache-write` input
- `src/cleanup-java.ts` — early return in `saveCache()` when `cache-write` is `false`
- `dist/` — rebuilt

Same change going into setup-node, setup-python, setup-go, setup-dotnet.


brunoborges:
After reviewing the current GitHub Actions cache security model, I do not think this input is needed for the stated cache-poisoning threat.

- Caches created by `pull_request` runs are scoped to `refs/pull/.../merge`; they cannot be restored by the base branch or other pull requests.
- Low-trust events that resolve to the default branch, including `pull_request_target`, `issue_comment`, and `workflow_run`, receive read-only cache access from the backend.
- setup-java now ships `@actions/cache` 6.2.0, which honors `ACTIONS_CACHE_MODE` and skips saves when writes are disallowed.

The proposed example, `cache-write: ${{ github.event_name != 'pull_request' }}`, therefore disables useful PR-local cache saves without protecting `main`; it also evaluates to true for `pull_request_target`, where backend enforcement is the actual protection.

A restore-only option could still be considered separately as an ergonomic feature, but it is not required as a security control. GitHub recommends `actions/cache/restore` when a workflow explicitly needs restore-only behavior.

References:
- https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache
- https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#cache-access-for-low-trust-workflow-triggers
- https://github.com/actions/setup-java/pull/1128

Closing this PR as unnecessary for the stated problem.

PR #553, opened 2023-11-07
Add a toolchain only - control update of settings.xml, toolchains.xml, JAVA_HOME and PATH more granular

mhoffrog:
09-Nov-2023: Amended commit comments only.

**Description:**
The following input items have been added (copied from README.md update):
  - `update-env-javahome`: By default action updates `env.JAVA_HOME` with the path of java installed. In order to skip update of JAVA_HOME, set this to `false`. The creation of the env variable JAVA_HOME_{{ MAJOR_VERSION }}_{{ ARCHITECTURE }} is NOT affected by this item. That will be created for any setup.

  - `update-env-path`: By default action adds `<java_install_dir>/bin` to `env.PATH`. In order to skip this, set this to `false`.

  - `update-toolchains-only`: If set to true, then `overwrite-settings`, `update-env-javahome` and `update-env-path` are propagated to `false` if the specific one is not explicitly configured to `true`. Only a toolchain entry will be added to toolchains.xml. Default is `false`.

**Summary on a few code improvements** (Inserted by 09-Nov-2023):
By that implementation I did a few code improvements as follows:
- `setup-java.ts`:
  - renamed interface installerInputsOptions -> IInstallerInputsOption and moved the definition to the top of the source file to levarage readability
- input values for `overwrite-settings` - plus the ones I did add - are read now once only in a single code place (within `setup-java.ts` only) to a constant and pass their effective values either by method parameters or by object fields to other places during the runtime lifecycle
- variable naming within `toolchains.ts`
- minor improvements of wording on log messages regarding toolchains.xml updates.

Further particular details on the code changes can be taken from my commit comment https://github.com/actions/setup-java/commit/b0dadd279479dcb798d4d5ab673fbd3c27b7fb0b.

**Related issue:**
#552

**Check list:**
- [x] Mark if documentation changes are required.<br>:bulb: **Documentation has been updated by this PR**
- [ ] Mark if tests were added or updated to cover the changes.

Issue #552, opened 2023-11-06
Allow to add a toolchain only - without updating settings.xml, JAVA_HOME and PATH. Only toolchains.xml must be extended.


mhoffrog:
**Description:**
Allow to add another toolchain only by a subsequent setup-java - only toolchains.xml must be updated with the additional JDK setup and the JDK specific JAVA_HOME_<jdk version> must be set.
No update of settings.xml, no add to PATH, no update of JAVA_HOME.

**Justification:**
In my scenario I run a first setup-java configuring settings.xml and setting up my JAVA_HOME JDK.
Then I determine a probable additionally required toolchain version from the compile target setting of my Maven projects POM.
If this is requiring a different JDK, then I am running another setup-java for this toolchain.
In this case I want to get the toolchains.xml being extended only and to preserve JAVA_HOME, PATH and settings.xml from my first setup-java.
Currently I am working around this requirement by a step preserving JAVA_HOME, PATH and settings.xml before the setup-java step and restoring those after that step. This is working fine so far.
Another config property for setup-java like e.g. `add-toolchain-only` would save those efforts!

**Are you willing to submit a PR?**
Yes - cannot promise a timeline yet.

IvanZosimov:
Hi, @mhoffrog 👋 Thank you for the feature request, we will take a look and get back to your with our decision.

TWiStErRob:
I would like this too, use case similar to https://github.com/orgs/community/discussions/73309

Pretty much anyone who provides a composite action that runs a jar or JDK commands is affected. I have no idea how others do it, probably delegate the labor to every user 😞.

mhoffrog:
@TWiStErRob I did file PR #553 already and did keep it up to date with recent master - s. https://github.com/actions/setup-java/pull/553#issuecomment-1919726093.
Unfortunately I do not get any further feedback nor answer on my comments on this PR - its a bit frustrating just getting no reaction for the time being.
@IvanZosimov Maybe you are all pretty busy - would there be a person in the team to assign for review or to put attention to this item?

caquino:
Hi, we are facing a similar issue: our self-hosted runners already have `settings.xml` pre-configured, as we use configurations much more complex than what `actions/setup-java` supports. However, when `overwrite-settings: false`, it also affects `toolchains.xml` management.
Ideally, we should have a different setting like `overwrite-toolchains: false` as these configurations have different requirements and controls.

mhoffrog:
[...]

PR #1017, opened 2026-06-12
feat: Add set-default option

### LINKED ISSUE #560: Input property that would configure if the JDK should be configured as default or not
**Description:**

Currently, `setup-java` configures the installed Java as default, however, there are cases when no change in `JAVA_HOME` is needed.

I suggest adding `configure-as-default` boolean input field (default `true`) so users could opt-out of configuring the installed Java for the default one.

**Justification:**

https://github.com/oracle-actions/setup-java downloads java binary, and it then calls `actions/setup-java` to set it up.

I use `oracle-actions/setup-java` to fetch Java EA (e.g. Java 22). I do not want Java 22 to be set as default since it would fail the build (e.g. Gradle does not yet support running with Java 22).
I need Java 22 for tests only, so I want Java 22 to be present, however, I would like to keep `JAVA_HOME` untouched.

Maybe

gsmet:
**Description:**

Add a `set-default` option allowing to install an additional JDK without making it the default one.

I have wanted this for quite a long time as I'm running custom GitHub Actions with Java, which might require a specific JDK and I don't want to pollute the JDK that is used by the overall workflow calling the action.
And I'm apparently not alone as there was a preexisting issue.

**Related issue:**

Fixes #560

**Check list:**
- [X] Mark if tests were added or updated to cover the changes.


brunoborges:
Cross-linking for coordination: this `set-default` option overlaps conceptually with the older PR #553 (issue #552), which also lets you install a JDK without mutating `JAVA_HOME`/`PATH`.

`set-default: false` here is effectively a cleaner, single-flag expression of #553's `update-env-javahome: false` + `update-env-path: false` (both also deliberately preserve `JAVA_HOME_<major>_<arch>`). We're treating **`set-default` as the canonical primitive** for "don't make this JDK the default."

The one piece #553 covered that this PR does **not** is a **Maven toolchains-only** install (register the JDK in `toolchains.xml` only). That remaining gap has been split out into #1047, scoped to build **on top of** `set-default` rather than re-implement env skipping. PR #553 is being closed in favor of this PR + #1047.

No change requested here — just establishing the lineage: #560 (this PR) and #552 → #1047.

gsmet:
@brunoborges thanks for taking good care of this PR. Let me know if you need anything from me.

brunoborges (approved):

brunoborges (approved):

gsmet:
Awesome! Thanks for merging!

PR #1045, opened 2026-06-22
chore: upgrade ESLint to v10 with flat config

brunoborges:
## Summary

Upgrades ESLint from **8.57.0 → 10.5.0**. ESLint 10 removes support for the legacy `.eslintrc` format entirely, so this migrates the configuration to the new **flat config** (`eslint.config.mjs`).

This supersedes the standalone Dependabot bump in #1022, which could not work on its own: bumping `eslint` to 10 without coordinating the `@typescript-eslint` upgrade and the flat-config migration triggers an `ERESOLVE` peer conflict (the old `@typescript-eslint/eslint-plugin` caps `eslint` at `^9`) and leaves the action with an unsupported `.eslintrc.js`.

## Changes

- Bump `eslint` `8.57.0` → `10.5.0` and `eslint-plugin-jest` → `29.15.2`.
- Replace the legacy `@typescript-eslint/eslint-plugin` + `@typescript-eslint/parser` pair with the `typescript-eslint` meta-package (`8.62.0`), which supports ESLint 10.
- Replace the deprecated, unmaintained `eslint-plugin-node` (broken on ESLint 9+) with its maintained fork `eslint-plugin-n` (`18.1.0`). The only rule in use, `node/no-extraneous-import`, becomes `n/no-extraneous-import`.
- Add `@eslint/js` (`10.0.1`) and `globals` (`17.7.0`), now required by flat config.
- Translate `.eslintrc.js` → `eslint.config.mjs` and `.eslintignore` → the flat-config `ignores`, preserving **all** existing rules and the test-file overrides.
- Update the `lint` / `lint:fix` scripts to invoke `eslint` directly (flat config is auto-detected; the `--config ./.eslintrc.js` flag is removed).

## Behavior preservation

- `preserve-caught-error` (new in ESLint 10's `recommended` set) is explicitly disabled to keep the previous lint behavior. Adopting it requires attaching an `Error` `cause` (ES2022 target) and is left as a follow-up.
- **Only dev/lint dependencies change.** Runtime dependencies are frozen at their `main` versions, so the built `dist/` output is byte-for-byte unchanged (no `dist/` diff in this PR).

## Validation

- ✅ `npm ci` reproduces the lockfile
- ✅ `npm run lint` (ESLint 10 flat config) passes clean
- ✅ `npm run format-check` passes
- ✅ `npm run build` — `dist/` identical to `main`
- ✅ `npm test` — 762/763 (the one failure is a sandbox network error unrelated to this change)
- ✅ `npm audit --audit-level=high` — no high-severity advisories

## Note on auto-sync

`.eslintrc.js` was previously auto-synced from [actions/reusable-workflows](https://github.com/actions/reusable-workflows). Upstream has not yet published a flat config, so this migration intentionally deviates from that auto-sync until upstream catches up.


gdams:
I had a quick look into this one myself, I was concerned that there was some automated tooling in place to sync from [actions/reusable-workflows](https://github.com/actions/reusable-workflows). If that's the case this would break this PR. I wonder if it's worth raising a fix in that repo first?

gdams:
superceded by https://github.com/actions/setup-java/pull/1078

PR #1053, opened 2026-06-23
feat: suppress Maven transfer progress via MAVEN_ARGS by default (add show-download-progress input)

brunoborges:
## Description

By default, Maven logs a line for every artifact it downloads, which adds significant noise to CI logs. This PR makes `setup-java` set the [`MAVEN_ARGS`](https://maven.apache.org/configure.html#maven_args-environment-variable) environment variable to include `-ntp` (`--no-transfer-progress`) so that Maven invocations in the job produce cleaner, smaller logs — without needing to add `-ntp` to every `mvn` command.

A new optional input lets users opt back into the progress output.

## Behavior

- **Default (`show-download-progress: false`)**: the action exports `MAVEN_ARGS` with `-ntp` appended.
  - Any **existing `MAVEN_ARGS`** value is preserved (the flag is appended, not overwritten).
  - **Idempotent**: the flag is not added if `MAVEN_ARGS` already contains `-ntp` or `--no-transfer-progress`.
- **Opt-in (`show-download-progress: true`)**: `MAVEN_ARGS` is left untouched, preserving Maven's default progress output and any user-provided value.
- Applies on **all platforms**. `MAVEN_ARGS` is honored by **Maven 3.9.0+** and the Maven Wrapper; older Maven versions ignore it (no-op). It has **no effect on non-Maven builds** (Gradle, sbt, etc.).

## Example

```yaml
- uses: actions/setup-java@v5
  with:
    distribution: 'temurin'
    java-version: '21'
    show-download-progress: true # keep Maven download/transfer progress in the logs
```

## Changes

- **`action.yml`**: add `show-download-progress` input (default `false`).
- **`src/constants.ts`**: add the input name and `MAVEN_ARGS` / flag constants.
- **`src/maven-args.ts`**: new `configureMavenArgs()` implementing the logic above.
- **`src/setup-java.ts`**: invoke `configureMavenArgs()` during setup.
- **`__tests__/maven-args.test.ts`**: unit tests (default, opt-in, existing-value preservation, idempotency for both flag spellings).
- **`docs/advanced-usage.md`**: new "Maven transfer progress" section + TOC entry.
- **`dist/`**: rebuild bundled action.

## Notes

- This is the correct mechanism for suppressing transfer progress, distinct from `<interactiveMode>false</interactiveMode>` (batch mode), which does **not** control progress output. See the discussion on #1052.

## Check list


brunoborges:
@copilot resolve the merge conflicts in this pull request

slawekjaranowski:
@brunoborges please also add new option to action documentation 😄

slawekjaranowski:
ok, there is at https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#Maven-transfer-progress-download-logs

will be easier to find to also add to:
https://github.com/actions/setup-java#maven-options

brunoborges:
Done.

cholmcc:
Note, this breaks f.ex. https://github.com/lenucksi/SieveEditor/blob/master/scripts/fetch-maven-deps-for-flatpak.sh which relies on the verbose output of `mvn`.  Perhaps the default for `show-download-progress` should be `true`. 

Yours,
_Christian_

brunoborges:
> Note, this breaks f.ex. https://github.com/lenucksi/SieveEditor/blob/master/scripts/fetch-maven-deps-for-flatpak.sh which relies on the verbose output of `mvn`. Perhaps the default for `show-download-progress` should be `true`.
> 
> Yours, _Christian_

Thanks @cholmcc, appreciate the feedback. I'll keep this in mind. However, I see this as an exceptional case.

PR #1070, opened 2026-07-01
Updated jetbrains test: https.request() now catches errors. This fixes leaking tests as well

jmjaffe37:
**Description:**
Now running https.request() within a try/catch to make sure that errors are caught and the test if rejected upon error. This fixes an issue where `npm run check` would always finish with the following warning:
`A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them`

Also removed deprecated code lines from `.husky/pre-commit` and `.husky/pre-push`. These lines would begin to fail in husky version 10.0.0, but this PR fixes that :)

**Related issue:**

**Check list:**

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,1 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
 npx lint-staged
--- a/__tests__/distributors/jetbrains-installer.test.ts
+++ b/__tests__/distributors/jetbrains-installer.test.ts
@@ -89,14 +89,20 @@ describe('findPackageForDownload', () => {
 
       await new Promise<void>((resolve, reject) => {
         const request = https.request(url, options, res => {
+          let assertionError: unknown;
+
           try {
             // JetBrains uses 403 for non-existent packages
             expect(res.statusCode).not.toBe(403);
-            res.resume();
-            res.on('end', resolve);
           } catch (error) {
-            reject(error);
+            assertionError = error;
           }
+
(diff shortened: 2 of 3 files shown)


__tests__/distributors/jetbrains-installer.test.ts
```diff
@@ -87,10 +87,20 @@ describe('findPackageForDownload', () => {
...
+      await new Promise<void>((resolve, reject) => {
+        const request = https.request(url, options, res => {
+          try {
+            // JetBrains uses 403 for non-existent packages
+            expect(res.statusCode).not.toBe(403);
+            res.resume();
+            res.on('end', resolve);
+          } catch (error) {
+            reject(error);
+          }
```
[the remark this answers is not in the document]
jmjaffe37: Applied

.husky/pre-commit
```diff
@@ -1,4 +1 @@
-#!/usr/bin/env sh
```
brunoborges, line 1: @jmjaffe37 what was the issue here that suggests the removal of this code block ?

brunoborges (changes requested):

.husky/pre-commit
jmjaffe37: @brunoborges, this line is deprecated as per the warnings that husky generates whenever there is a commit or a push. It says that these lines will cause failures starting on husky version 10. See the attached screenshot below

image

PR #1079, opened 2026-07-08
Map Zulu x86 architecture to i686 for Azul Metadata API

brunoborges:
## Summary

Follow-up to #1010 (Azul Metadata API migration). Fixes a regression where an explicit `architecture: x86` request for the Zulu distribution could resolve to a **64-bit** JDK.

## Problem

The migration mapped code arch `x86` → API `arch=x86`. However, the Azul Metadata API's `arch=x86` returns **both** 32-bit (`i686`) and 64-bit (`x64`) packages. The two variants share identical `java_version` and `distro_version`, so `findPackageForDownload` cannot distinguish them and picks whichever the API returns first.

Consequences for `architecture: x86`:
- It frequently resolves to a **64-bit** build instead of 32-bit.
- For Java 21+ (where Azul dropped 32-bit), it **silently returns a 64-bit JDK** instead of failing.

The legacy Zulu Discovery API avoided this by using `arch=x86&hw_bitness=32`, which returned only 32-bit builds.

## Fix

Map `x86` → `i686`. The Metadata API's `arch=i686` returns **only** genuine 32-bit builds, with verified full version parity to the legacy 32-bit behavior:

| Platform | legacy `x86` + `hw_bitness=32` | new `i686` | lost |
|---|---|---|---|
| linux | 141 versions | 141 | 0 |
| windows | 143 versions | 143 | 0 |

x64 and aarch64 resolution are unaffected.

## Testing

- Updated the three Zulu installer test suites (URL query params + `getArchitectureOptions` mapping).
- Full suite passes (855 tests).
- `dist/` rebuilt.

### DIFF
--- a/__tests__/distributors/zulu-installer.test.ts
+++ b/__tests__/distributors/zulu-installer.test.ts
@@ -87,7 +87,7 @@ describe('getAvailableVersions', () => {
         packageType: 'jdk',
         checkLatest: false
       },
-      '?os=macos&archive_type=tar.gz&java_package_type=jdk&javafx_bundled=false&crac_supported=false&arch=x86&release_status=ga&availability_types=ca&page=1&page_size=100'
+      '?os=macos&archive_type=tar.gz&java_package_type=jdk&javafx_bundled=false&crac_supported=false&arch=i686&release_status=ga&availability_types=ca&page=1&page_size=100'
     ],
     [
       {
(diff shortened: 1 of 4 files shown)


gdams:
Should we not include i386 in this logic too? Or does the Azul API specifically map to i686?

brunoborges:
Seems clear to me this is a matter of them mapping i686

PR #1088, opened 2026-07-08
feat: expose cache-primary-key output (#597)

### LINKED ISSUE #597: Cache integration between `actions/setup-java` and `actions/cache`/`actions/cache/restore`
**Description:**
Expose `cache-primary-key` as an output to allow integration with `actions/cache`and `actions/cache/restore`.

**Justification:**
Caching rules differ according to the build tool used (mvn/sbt/etc), and I would like to leverage the logic already implemented on `actions/setup-java` to handle the specifics for each build tool.

That being said, I have dependent jobs that I would like to guarantee that the same cache is used by forcing the same cache key.

I understand that any advanced cache usage should be done manually using `actions/cache`, but just exposing the cache key used by `actions/setup-java` as an output, like `cache-hit`, would make it possible to extend the functionality from `actions/setup-java`.

My understanding is that a small change [here](https://github.com/actions/setup-java/blob/7a445ee88d4e23b52c33fdc7601e40278616c7f8/src/cache.ts#L113) to set an output, and to `action.yml` to declare the output, would suffice.
I would be more than happy to submit a PR, but before doing so, I would like to check with the team if this aligns with the vision for the future.

brunoborges:
## What

Exposes the primary cache key computed by the caching logic as a new `cache-primary-key` action output.

Closes #597.

## Why

Today the action computes a primary cache key internally (`setup-java-<os>-<arch>-<packageManager>-<hash>`) but never surfaces it. Users who want to compose the built-in setup-java caching with [`actions/cache`](https://github.com/actions/cache) / [`actions/cache/restore`](https://github.com/actions/cache/tree/main/restore) — e.g. to force the same key across dependent jobs — have no way to reference it. This exposes it as an output, mirroring what `actions/setup-node` and `actions/cache` already do.

## Changes

- `src/cache.ts`: set the `cache-primary-key` output in `restore()`
- `action.yml`: declare the new `cache-primary-key` output
- `README.md`: document the new output
- `__tests__/cache.test.ts`: assert the output is set
- `dist/`: rebuild

## Notes

- Non-breaking, purely additive.
- The output is empty when caching is not enabled (the action only computes/sets it inside `restore()`, which runs only when `cache:` is set).
- A backport of this change to `releases/v5` is in a separate PR.

## Testing

- `npm test` (cache suite: 32/32 passing)
- lint + format-check clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ When the option `cache-dependency-path` is specified, the hash is based on the m
 
 The workflow output `cache-hit` is set to indicate if an exact match was found for the key [as actions/cache does](https://github.com/actions/cache/tree/main#outputs).
 
-The workflow output `cache-primary-key` exposes the primary cache key computed by the action for the configured build tool. It is useful for composing with [`actions/cache`](https://github.com/actions/cache) or [`actions/cache/restore`](https://github.com/actions/cache/tree/main/restore) in later steps or dependent jobs that need to reuse the exact same key. It is empty when caching is not enabled.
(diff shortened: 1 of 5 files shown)


brunoborges:
@copilot rebase with main, rebuild the project and push the updated dist folder to this PR

Issue #269, opened 2022-01-04
Gradle cache increases by 400 MB every time updating Gradle wrapper version


DevCharly:
**Description:**
When updating the Gradle wrapper to a new version (changed `gradle-wrapper.properties`), then
1. setup-java@v2 first restores the gradle cache from **previous run** (why??)
2. new Gradle version is downloaded and build
3. then setup-java@v2 creates a new cache file (~800 MB) which is about 400 MB larger than before because it contains two Gradle versions

Changing Gradle wrapper version again increases the cache file size by 400 MB to 1200 MB, and so on...

Here is a run where it happened: https://github.com/JFormDesigner/FlatLaf/runs/4705134891

In "Setup Java 17" the outdated cache is restored (https://github.com/JFormDesigner/FlatLaf/runs/4705134891#step:4:35):
~~~
Cache Size: ~425 MB (445413982 B)
/usr/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/19b74bcf-0264-4065-b1ef-448c93c2d4fc/cache.tzst -P -C /home/runner/work/FlatLaf/FlatLaf
Cache restored successfully
Cache restored from key: setup-java-Linux-gradle-707ed6f8250ea8ec2ee660640962308e354fe354d36338715ef1bcab4fcc837f
~~~

In "Post Setup Java 17" a new cache with ~836 MB is created (https://github.com/JFormDesigner/FlatLaf/runs/4705134891#step:11:5):
~~~
Cache Size: ~836 MB (876885346 B)
Cache saved successfully
Cache saved with the key: setup-java-Linux-gradle-af7e537f7a03362010edc6596ab7fba05bdfb901982028cbd2d21e5291cd932d
~~~

Note the different hash keys in the cache file names!

**Task version:**
actions/checkout@v2

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [x] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Repro steps:**  
See above description.

**Expected behavior:**
When changing `gradle-wrapper.properties` (or `build.gradle`), then the cache should be not restored.

**Actual behavior:**
Cache is restored even if `gradle-wrapper.properties` (or `build.gradle`) changed, which increases the size of the cache file by aboud 400 MB every time a new Gradle wrapper version is used.

dmitry-shibanov:
Hello @DevCharly. Thank you for your report. We'll investigate the issue.

[...]

PR #1097, opened 2026-07-10
Cache Maven and Gradle wrapper distributions separately from the dependency cache

### LINKED ISSUE #1095: Maven wrapper caching does not seem to cache enough
**Description:**

setup-java was recently modified (2 weeks ago) to also cache Maven distributions used by the `mvnw` wrapper script. See #1027.

JRuby's build uses `setup-java@v5` but we are still seeing failed downloads of the distribution. We would expect this to have been cached long ago, either by us or by other projects, so this is unexpected.

See a failure as recently as yesterday, associated with a PR to attempt to reduce these failures: https://github.com/jruby/jruby/actions/runs/28970521797/job/85964762963

**Task version:**

`setup-java@v5` as of the time of the above build.

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

I know we see it on Ubuntu, but I believe it has affected builds on other platforms as well. It does not appear to be specific to any platform.

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Repro steps:**  
A description with steps to reproduce the issue. If your have a public example or repo to share, please provide the link.

The build below shows a failed mvnw wrapper download that killed a JRuby CI job.

https://github.com/jruby/jruby/actions/runs/28970521797/job/85964762963

**Expected behavior:**

We use `cache: maven` and `setup-java@v5` so we expected that the wrapper would not be downloaded from Maven Central and would not trigger this error.

**Actual behavior:**

As others have reported, rate-limiting or similar restrictions at the Ruby Central level appear to be causing mvnw dist downloads to fail occasionally. We want to use setup-java caching to prevent re-downloading of this distribution, since it almost never changes. That does not appear to be happening.

brunoborges:
**Description:**

Each build tool's wrapper distribution was cached in the **same** cache entry as its dependency cache, keyed on volatile dependency files. Because those files change frequently and no `restoreKeys` are used (by design, see #269), almost every change rotates the key and produces a full cache miss. On a miss nothing is restored, including the rarely-changing wrapper distribution, so `./mvnw` / `./gradlew` re-download the build tool, which intermittently fails due to upstream rate limiting.

**Maven:** The wrapper distribution (`~/.m2/wrapper/dists`) shared the `~/.m2/repository` entry, keyed on `**/pom.xml` (plus wrapper properties and extensions). I confirmed the impact from the linked JRuby run: the `setup-java` step logged `maven cache is not found`, then `./mvnw` failed with `wget: Failed to fetch https://repo1.maven.org/.../apache-maven-3.9.14-bin.zip`. JRuby's `maven-wrapper.properties` changed only twice in about two years, while the repo has 31 `pom.xml` files that each rotate the shared key, so the wrapper was evicted far more often than it actually changes.

**Gradle:** The wrapper distribution (`~/.gradle/wrapper`) shared the `~/.gradle/caches` entry, keyed on volatile files such as `**/*.gradle*`, with the same net effect — the wrapper was re-downloaded on nearly every dependency change.

**Approach:**

Give each build-tool wrapper distribution its own cache entry, keyed only on its wrapper properties file, so it survives the frequent dependency changes that invalidate the main dependency cache.

- Adds a generic `additionalCaches` concept to `PackageManager`, restored and saved independently with name-scoped state keys so restore/save state does not collide with the main cache.
- Moves `~/.m2/wrapper/dists` out of the main maven `path` into a `maven-wrapper` additional cache keyed on `**/.mvn/wrapper/maven-wrapper.properties`.
- Moves `~/.gradle/wrapper` out of the main gradle `path` into a `gradle-wrapper` additional cache keyed on `**/gradle-wrapper.properties`.
- Each additional cache is skipped silently when the project does not use the corresponding wrapper (no wrapper properties file).
- Keeps the `cache-hit` and `cache-primary-key` outputs driven by the main dependency cache, so existing behavior and outputs are unchanged.

**Related issue:**

Fixes: #1095

**Check list:**


brunoborges:
Backport PR #1122

PR #1109, opened 2026-07-13
feat: Add input option add-toolchain-only

### LINKED ISSUE #1047: Add a "toolchains-only" install mode (layered on the set-default primitive)
### Background

PR #1017 introduces a `set-default` input: when `set-default: false`, the action installs a JDK **without** updating `JAVA_HOME` / `PATH` (while still exporting `JAVA_HOME_<major>_<arch>`). That cleanly covers the "don't make this JDK the environment default" use case (issue #560).

The earlier PR #553 (issue #552) aimed at a broader, Maven-centric goal that **partially overlaps** with `set-default` but also contains a piece that `set-default` does **not** cover. This issue captures that remaining gap so it can be built **on top of** the `set-default` primitive rather than as a competing implementation.

### The gap not covered by `set-default`

`set-default: false` controls the **environment** (`JAVA_HOME` / `PATH`). It does not, on its own, add the installed JDK as a **Maven toolchain** entry. The use case from #552 / #553 is:

> Install a JDK and register it **only** as a Maven toolchain (`~/.m2/toolchains.xml`) — without touching `JAVA_HOME`, `PATH`, or `settings.xml`.

This is useful when a workflow needs an additional JDK available to Maven's `maven-toolchains-plugin` (e.g. compiling/running a module against a specific JDK) while leaving the workflow's primary Java environment untouched.

### Proposed scope (narrowed)

Add a focused input — e.g. `add-toolchain-only` (name TBD) — that, when `true`:

- [ ] Adds/updates the JDK entry in `toolchains.xml` (existing toolchains logic).
- [ ] Implies `set-default: false` semantics (does **not** update `JAVA_HOME` / `PATH`) unless the user explicitly opts back in.
- [ ] Does **not** modify `settings.xml` (no `overwrite-settings` side effects) unless explicitly requested.
- [ ] Still exports `JAVA_HOME_<major>_<arch>` (consistent with `set-default: false`).

Explicitly **out of scope** (now handled by #1017):
- General `JAVA_HOME` / `PATH` skipping — use `set-default: false`.

### Relationship / dependencies

- **Depends on / builds on:** #1017 (`set-default`) — should be implemented as a thin layer over that primitive, reusing its env-skip behavior instead of re-deriving it.
- **Supersedes (scope-wise):** the toolchains-only portion of PR #553.
- **Original request:** #552.

### Notes

PR #553 also proposed `update-env-javahome` / `update-env-path` as separate booleans; those are intentionally **not** carried forward — `set-default` from #1017 is the single, cleaner knob for that behavior. This issue deliberately narrows the remaining work to just the **toolchains.xml-only** capability.

mhoffrog:
**Description:**
In case of installing multiple Java versions, this option will add a subsequent installed JDK to the Maven toolchains.xml only without updating Maven settings.xml nor updating environment vars JAVA_HOME and PATH.

**Changes**
- Add input option `add-toolchain-only` - default: `false`
- Propagate default values for `set-default` and `overwrite-settings` from `add-toolchain-only`
- setup-java.ts:
  - couple of code improvements for less error prone interface instancing
  - single place for reading input of `overwrite-settings`
- toolchain.ts, auth.ts:
  - replace reading of input `overwrite-setting` by passing `overwriteSettings` and `addToolchain` as method parameters
- __tests__/toolchains.test.ts: followup previous changes
- documentation updated in `README.md` and `docs/advanced-usage.md`.

**Related issue:**
Closes #1047 - Followup Issue of #553

@brunoborges - many thanks for having created #1047.

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ For more details,  see the full release notes on the [releases page](https://git
 
   - `check-latest`: Setting this option makes the action to check for the latest available version for the version spec.
 
-  - `set-default`: Set to `false` to install a JDK without making it the default. When `false`, `JAVA_HOME` and `PATH` are not updated, but `JAVA_HOME_<major>_<arch>` is still set so the JDK remains discoverable. Default value: `true`. See [Installing JDK without setting as default](docs/advanced-usage.md#Installing-JDK-without-setting-as-default) for more details.
+  - `set-default`: Set to `false` to install a JDK without making it the default. When `false`, `JAVA_HOME` and `PATH` are not updated, but `JAVA_HOME_<major>_<arch>` is still set so the JDK remains discoverable. Default value: `!add-toolchain-only`. If explcitly set `true`, it will update JAVA_HOME and PATH regardless of `add-toolchain-only`. See [Installing JDK without setting as default](docs/advanced-usage.md#Installing-JDK-without-setting-as-default) for more details.
 
   - `verify-signature`: Verifies downloaded Java package signatures when supported by the selected distribution. Currently supported for `temurin` and `microsoft`. If set to `true` for unsupported distributions, the action fails.
 
@@ -58,7 +58,9 @@ For more details,  see the full release notes on the [releases page](https://git
   #### Maven options
   The action has a bunch of inputs to generate maven's [settings.xml](https://maven.apache.org/settings.html) on the fly and pass the values to Apache Maven GPG Plugin as well as Apache Maven Toolchains. See [advanced usage](docs/advanced-usage.md) for more.
 
-  - `overwrite-settings`: By default action overwrites the settings.xml. In order to skip generation of file if it exists, set this to `false`.
+  - `overwrite-settings`: Overwrite the settings.xml file if it exists. Default is `!add-toolchain-only`. If explcitly set `true`, it will update settings.xml regardless of `add-toolchain-only`.
+
+  - `add-toolchain-only`: Update toolchains.xml only. Default is `false`. No update of settings.xml, no update of JAVA_HOME and PATH unless `overwrite-settings` or `set-default` are explicitly set `true`.
 
   - `server-id`: ID of the distributionManagement repository in the pom.xml file. Default is `github`.
 
--- a/__tests__/toolchains.test.ts
+++ b/__tests__/toolchains.test.ts
@@ -87,7 +87,7 @@ describe('toolchains tests', () => {
     await toolchains.createToolchainsSettings({
       jdkInfo,
       settingsDirectory: altHome,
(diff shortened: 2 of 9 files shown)


mhoffrog:
Put to draft now to await merge of #1111.

brunoborges:
@mhoffrog I am not sure about this new feature. I think it adds complexity. What is the downside of having more JDKs installed with their specific environment variables configured? What is the real point in adding JDKs to the tool chain but not to settings.xml ?

mhoffrog:
> @mhoffrog I am not sure about this new feature. I think it adds complexity. What is the downside of having more JDKs installed with their specific environment variables configured? What is the real point in adding JDKs to the tool chain but not to settings.xml ?

Typically you have one JDK being your default JDK and you have to setup settings.xml only once. settings.xml does not know anything about JDKs installed. Settings.xml is about providing credentials and repository URLs. Multiple JDKs in toolchains.xml is typical scnario, since your POMs will define the toolchain for compiling and any plugin looking up toolchain config. Your Maven build job may have to run by another JDK, since Maven plugins have their own requirement to run on. Maven toolchain is to separate compile and unit testing JDks from Maven build job executing JDK.

Similar for Gradle - you will run your Gradle build job with JDK A and you may have unit and integration testing demands with a matrix of JDKs to perform.

> What is the downside of having more JDKs installed with their specific environment variables configured?

This is exactly what is to be avoided. There must be only a single environment var JAVA_HOME and PATH setting over all. Toolchains is to capture any other JDK required by your build job. This is less complexity at the end.

brunoborges:
Ok, but what is the problem with the current capabilities? You do get different JDKs configured in the toolchains for the example you gave.

mhoffrog:
> Ok, but what is the problem with the current capabilities? You do get different JDKs configured in the toolchains for the example you gave.

The current behavior is that with each additional JDK installed settings.xml will be overwritten. You can disable this by disabling overwriting, but then also toolchains will not get added the additional JDK. In addition you may disable set-default - but you have to do this for each particular additional JDK setup. This is more config hell than otherwise.

mhoffrog:
@brunoborges This PR may become obsolete, in case #1111 will have been merged. With #1111 toolchains will be updated/enhanced in any case. By that given, `add-toolchain-only` will be adequate to:
- `set-default: false` and
- `overwrite-settings: false`

So there is no need to add another input for that. The only use case not covered anymore will be to disable toolchains update.

Please let me know, if this is as intended, then I will close this PR accordingly.

mhoffrog:
Closing this PR since use case can be resolved as described above. If there is need to disable toolchains update for some reason, a new feature request should be created.

PR #1111, opened 2026-07-13
Preserve Maven toolchains across repeated setup-java runs (#1099)

### LINKED ISSUE #1099: Further executions of setup-java overwrite previous toolchain entries for different JDK versions
**Description:**
Previously, it was possible to execute setup-java multiple times with different JDKs or JDK versions and have them all in the Maven toolchains XML. Now, it is always overwritten (contrary to the documentation under https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#Modifying-Maven-Toolchains).

Using one execution of setup-java with multiple versions still works, but has the drawback that the output contains only the path of the last version.

**Task version:**
Since v5.5.0, probably because of #534.

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [ ] Hosted
- [x] Self-hosted

**Repro steps:**  
Execute setup-java consecutively with different versions. Output the Maven toolchains XML.

**Expected behavior:**
Multiple toolchain entries should be present.

**Actual behavior:**
Only a single toolchain entry is present.

brunoborges:
## Summary

Fixes #1099. Running `setup-java` multiple times in the same job (matrix builds or multiple `java-version` values) could drop toolchain entries registered by earlier runs, leaving only one JDK in `~/.m2/toolchains.xml`.

## Root cause

Toolchains generation was gated behind the `overwrite-settings` input. When `overwrite-settings: false`, `writeToolchainsFileToDisk` skipped writing entirely if the file already existed — so a second `setup-java` run never persisted its JDK, silently discarding the merged result.

This gating was inappropriate because:
- `overwrite-settings` is documented (in `action.yml`) to control regeneration of **`settings.xml`** only.
- `generateToolchainDefinition` already performs a **non-destructive merge**: existing JDK, custom, and user-managed toolchains are preserved, and only an entry with the exact same `type` + `provides.id` is replaced.

So writing the toolchains file is always safe and should not be blocked.

## Changes

- **`src/toolchains.ts`**: Remove `overwriteSettings` from `configureToolchains`, `createToolchainsSettings`, and `writeToolchainsFileToDisk`. The toolchains file is now always written (non-destructively). `settings.xml` behavior in `auth.ts` is unchanged.
- **`__tests__/toolchains.test.ts`**: Update call sites, rewrite the "does not overwrite" test to assert non-destructive extension, and add a regression test that runs `configureToolchains` three times and asserts every JDK is preserved with no duplicates.
- **`docs/advanced-usage.md`**: Clarify that the merge is non-destructive and independent of `overwrite-settings`.
- **`dist/setup/index.js`**: Rebuilt.

## Testing

- `npm run build`, `npm run format-check`, `npm run lint` all pass.
- `npm test`: 909 passing, including the new regression test.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/__tests__/toolchains.test.ts
+++ b/__tests__/toolchains.test.ts
@@ -86,8 +86,7 @@ describe('toolchains tests', () => {
 
     await toolchains.createToolchainsSettings({
       jdkInfo,
-      settingsDirectory: altHome,
-      overwriteSettings: true
+      settingsDirectory: altHome
     });
 
     expect(fs.existsSync(m2Dir)).toBe(false);
@@ -135,8 +134,7 @@ describe('toolchains tests', () => {
 
     await toolchains.createToolchainsSettings({
       jdkInfo,
-      settingsDirectory: m2Dir,
-      overwriteSettings: true
+      settingsDirectory: m2Dir
     });
 
     expect(fs.existsSync(m2Dir)).toBe(true);
@@ -217,8 +215,7 @@ describe('toolchains tests', () => {
 
     await toolchains.createToolchainsSettings({
       jdkInfo,
-      settingsDirectory: m2Dir,
-      overwriteSettings: true
+      settingsDirectory: m2Dir
     });
 
     expect(fs.existsSync(m2Dir)).toBe(true);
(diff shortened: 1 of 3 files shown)


brunoborges:
@lmvysakh can you please review this?

mhoffrog:
@brunoborges This PR will raise a conflict to my #1109 one I guess.
=> #1109 will have to await this one being merged. I will take care of updating myone after this one has been merged.
=> I will put #1109 to draft now and will re-wake it later.

mahabaleshwars (approved):

lmvysakh (approved):

v-lmvysakh:
Hy @brunoborges ,

Reviewed and Approved PR.

PR #1126, opened 2026-07-15
Use YAML anchors to reduce boilerplate in e2e-versions workflow

brunoborges:
## Summary

Pure DRY/formatting refactor of `.github/workflows/e2e-versions.yml` — **no change to behavior, coverage, matrix values, step order, or names**. Once the YAML anchors are expanded, the document is byte-for-byte equivalent to the original.

## What changed

- **Checkout step** — the identical `Checkout` step (`actions/checkout@v7` with `persist-credentials: false`) was repeated 17 times. Anchored at its first occurrence with `&checkout_step` and replaced the other 16 with the alias `*checkout_step`.
- **Common OS matrix list** — `[macos-latest, windows-latest, ubuntu-latest]` was repeated 12 times. Anchored at its first occurrence with `&default_os` and aliased (`*default_os`) in the other 11 jobs that use that exact list. Lists that differ (e.g. containing `macos-15-intel`, `[ubuntu-latest]`, or `[windows-latest, ubuntu-22.04]`) were left untouched.

Anchors are defined **inline at their first real usage** (no top-level definitions block), since GitHub Actions rejects arbitrary top-level keys. YAML anchors are resolved by the parser before Actions schema validation, so inline `&`/`*` is valid.

Net: `30 insertions(+), 77 deletions(-)`.

## Validation

- Loaded the file with a YAML parser that expands anchors and confirmed the expanded document is **equal to the original** (`new == orig`), still 17 jobs, each retaining its Checkout / setup-java / verify steps.
- Ran `actionlint`: it reports only pre-existing schema notices (matrix `version: 8`, `matrix.version`/`matrix.architecture` expression properties) that exist independently of this change — no anchor/alias errors were introduced.


brunoborges:
@copilot resolve the merge conflicts in this pull request

PR #1131, opened 2026-07-15
docs: update setup-java examples

HarithaVattikuti:
Updated the documentation examples in README.md and docs/advanced-usage.md to use actions/checkout@v7 and actions/setup-java@v6. Also removed the obsolete V2 vs V1 section from the README.

Validation: docs-only change.


brunoborges:
Hey @HarithaVattikuti good finding! Just wondering if we should time this after the v6 release? Should we push that release by end of the month? I'd like to push v5.6 before that.

v-HarithaVattikuti:
> Hey @HarithaVattikuti good finding! Just wondering if we should time this after the v6 release? Should we push that release by end of the month? I'd like to push v5.6 before that.

Thanks @brunoborges  — that makes sense.  Since this PR is specifically for v6 docs, I think we should merge it alongside the v6 release. Let’s push v5.6 first, then merge this closer to the v6 release

brunoborges (approved):

SimonMarquis:
👋 Hi @HarithaVattikuti and @brunoborges, this update broke the README.md rendering for `Usage` and `Maven options`, and adds unnecessary spacing in all touched files. Can we revert this change and only update the necessary parts?

PR #1210, opened 2026-08-05
Fix JDK resolution cache platform identity

### LINKED ISSUE #1209: JDK resolution cache key does not distinguish Linux glibc and musl artifacts
**Description:**

The JDK resolution-cache identity does not include the effective Linux libc/platform variant.

The cache key currently includes runner OS, distribution, package type, architecture, version specification, and stability:

https://github.com/actions/setup-java/blob/ab597f914a6894678251803485c09f413263b889/src/jdk-resolution-cache.ts#L205-L215

However, artifact resolution additionally depends on whether the runner uses Alpine/musl:

- Temurin resolves `alpine-linux` instead of `linux`:
  https://github.com/actions/setup-java/blob/ab597f914a6894678251803485c09f413263b889/src/distributions/temurin/installer.ts#L269-L283
- SapMachine resolves `linux-musl` instead of `linux`:
  https://github.com/actions/setup-java/blob/ab597f914a6894678251803485c09f413263b889/src/distributions/sapmachine/installer.ts#L237-L251

Both environments report `RUNNER_OS=Linux`. Therefore, identical setup-java inputs on glibc and musl runners generate the same resolution-cache identity even though they require incompatible artifacts.

A fresh cache hit is returned without invoking the distribution resolver:

https://github.com/actions/setup-java/blob/ab597f914a6894678251803485c09f413263b889/src/distributions/base-installer.ts#L290-L296

As a result, a glibc runner can restore an Alpine/musl release URL and checksum, or an Alpine runner can restore a glibc release.

This is especially easy to encounter in mixed Linux runner fleets. Concurrent matrix jobs can also race to reserve the same immutable weekly cache key, making the artifact used by later runs depend on which platform saved the entry first.

**Task version:**

`actions/setup-java@ab597f914a6894678251803485c09f413263b889` (`main`, unreleased v6 development)

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

The other side of the collision is Alpine Linux/musl.

**Runner type:**
- [ ] Hosted
- [x] Self-hosted

This requires Linux runners sharing the same repository/branch cache scope and a compatible cache path/version, such as a consistently configured ephemeral runner fleet.

**Repro steps:**

1. Configure two clean Linux x64 runners in the same repository and branch cache scope:
   - one Alpine/musl runner;
   - one Ubuntu/glibc runner.
2. Ensure the runners have compatible `RUNNER_TEMP` layouts so they use the same Actions cache version.
3. Run the Alpine job first and let its post-action save the caches.
4. Run the Ubuntu job with the same setup-java inputs.

Example workflow, assuming suitable self-hosted labels:

```yaml
jobs:
  seed-from-alpine:
    runs-on: [self-hosted, linux, x64, alpine]
    steps:
      - uses: actions/setup-java@ab597f914a6894678251803485c09f413263b889
        with:
          distribution: temurin
          java-version: '21'
          architecture: x64
          cache-jdk: true
      - run: java --version

  restore-on-glibc:
    needs: seed-from-alpine
    runs-on: [self-hosted, linux, x64, glibc]
    steps:
      - uses: actions/setup-java@ab597f914a6894678251803485c09f413263b889
        with:
          distribution: temurin
          java-version: '21'
          architecture: x64
          cache-jdk: true
      - run: java --version
```

The second job computes the same resolution-cache key and can restore the Alpine Temurin release. The restored JDK may then fail because the glibc system does not provide the musl dynamic loader.

Using `cache: maven`, `cache: gradle`, or `cache: sbt` without explicitly setting `cache-jdk: false` can also activate this path because dependency caching implicitly enables JDK caching.

**Expected behavior:**

Resolution-cache entries should only be shared when the complete resolver context is equivalent. Linux glibc and musl requests should produce different cache identities and restore compatible JDK artifacts.

**Actual behavior:**

Linux glibc and musl requests with otherwise identical inputs produce the same resolution-cache identity. A fresh cache entry can therefore bypass vendor resolution and select an artifact for the wrong libc.

**Suggested fix:**

- Add a canonical platform/libc discriminator, such as `linux-glibc` or `linux-musl`, to `JdkResolutionRequest`.
- Include that discriminator in `getResolutionIdentity()`.
- Increment `JDK_RESOLUTION_KEY_VERSION` so existing ambiguous entries are not restored.
- Add unit tests proving glibc and musl requests produce different identities.
- Add an Alpine-to-glibc cache E2E test for Temurin and SapMachine.

jdubois:
## Description

The JDK resolution cache currently treats all Linux runners with the same declared inputs as equivalent. Temurin and SapMachine resolve different artifacts on Alpine, however, so an Alpine/musl runner and a glibc runner can restore the same cached release metadata and install an incompatible JDK.

This change:

- adds a canonical runtime platform identity to every resolution-cache request;
- distinguishes `linux-glibc` from `linux-musl`;
- shares Alpine detection with the Temurin and SapMachine resolvers so the identity matches artifact selection;
- bumps the resolution-cache key version to avoid restoring ambiguous v1 entries;
- adds regression coverage for libc-specific key separation and platform normalization.

Closes #1209.

## Testing

- Full repository validation passed in the fork: 85 checks across Linux, macOS, and Windows.
- Basic validation passed, including formatting, linting, build, unit tests, and package audit.
- Cache, smoke, local-file, publishing, CodeQL, licensing, and zizmor workflows passed.
- A clean `check-dist` rebuild exactly matched the committed generated bundles.


brunoborges:
I reviewed this against a fix I'd written independently for #1209 and ended up preferring your approach, so rather than open a competing PR I've pushed one test-only commit here (3b8f519). Happy to drop it if you'd rather keep the branch yours.

### What the commit adds

Two gaps in `platform-types.ts` coverage:

- `getJavaPlatformIdentity`'s `normalizePlatform(platform) ?? platform` fallback had no test — the `it.each` covered only `linux`/`darwin`/`win32`, which all hit the alias map. Added `sunos → solaris` (alias path) and `aix → aix` (fallback path).
- `isAlpineLinux` had no direct test, so its `platform === 'linux'` short-circuit was unpinned.

Both verified by mutation rather than by eyeballing coverage:

| Mutation | Before | After |
|---|---|---|
| `?? platform` → `?? 'windows'` | 37 passed, 0 failed | fails |
| drop `platform === 'linux' &&` from `isAlpineLinux` | 37 passed, 0 failed | fails |

The short-circuit one is worth pinning beyond coverage bookkeeping: it's what keeps the `/etc/alpine-release` probe from running on non-Linux runners, so a stray file can never make Windows or macOS resolve as musl.

Full suite is green (41 suites / 1322 tests), format-check and lint clean. Test-only, so no `dist/` rebuild.

### On the design

My version used a `protected getResolutionPlatform()` hook on `JavaBase` that each distribution overrode to return its own `getPlatformOption()`. I talked myself into it on the grounds that it can't drift from artifact selection — but checking it against this branch, **yours is the better design and the hook is unnecessary.**

The hook only protects distributions that remember to override it, so its failure mode is a silent cross-libc collision — the exact bug in #1209. Yours is correct by default for every distribution, including ones that don't participate. I also couldn't find a single distribution that would need to override it today: identity is a function of (platform, arch, packageType, versionSpec, stable), and I checked the one candidate I suspected — Temurin's `includeJmods` — which derives from `packageType` and is therefore already covered. Even if Dragonwell/Corretto were later fixed to select `alpine-linux` artifacts, your identity already separates those correctly. So it'd be an abstraction with zero users.

Your injectable-parameter signature is also strictly better for testing than what I had, which mutated `process.platform` and spied on `fs`.

### Two optional notes

**The `v1 → v2` bump is redundant**, if you'd prefer to avoid the test churn. The restore key is `...-v{N}-{os}-{arch}-{digest}-` and the cache path embeds the digest too, so adding `platform` to the hashed object already changes the digest — no v1 entry can match the new prefix, and `@actions/cache`'s path-derived version check is a second layer. Harmless as intent-signalling, just not load-bearing. Maintainer call.

**Minor:** `expectedRequest` in `base-installer.test.ts` builds its expectation by calling `getJavaPlatformIdentity()`, the same function the production path calls. It still proves the wiring (a missing or wrong field fails), and the helper's behaviour is well covered by the contract tests, so this is fine as-is — just flagging that the assertion can't catch a regression inside the helper.

### Out of scope, for the record

Dragonwell and Corretto publish `alpine-linux` artifacts but their `getPlatformOption()` returns plain `linux` on Alpine, so they resolve glibc artifacts on musl runners. That's a pre-existing artifact-*selection* bug, unrelated to cache identity, and this PR neither causes nor worsens it. Worth a separate issue.

brunoborges (approved):
Approving as maintainer. Disclosure: I pushed two commits to this branch — a test-coverage commit for the platform-identity helper, and a fix carrying `platform` into the floating resolution request that #1219 introduced (the textual merge with main left the tree not compiling, and `dist/` stale).

Design and the core fix are jdubois'. Verified locally: 41 suites / 1365 tests, format-check, lint, and a clean `dist/` rebuild. All 84 CI checks green after re-running one unrelated Windows GPG-cleanup flake.

brunoborges:
Status: this is merge-ready — all 84 checks green (one unrelated Windows GPG-cleanup flake re-run), branch up to date with main, and I've approved it.

It can't actually merge yet, though: the `Copilot Code Review` ruleset sets `require_last_push_approval: true`, and I pushed the last commit (3fc4744). That commit is a production change, not just tests — merging main brought in #1219, whose `getFloatingResolutionRequest` predates the required `platform` field, so the merged tree didn't compile and `dist/` was stale.

So it needs an approving review from someone who didn't make that push — @jdubois or another maintainer. I'd rather leave it for a second pair of eyes than bypass the rule, especially since my commit touches `base-installer.ts`.

PR #1213, opened 2026-08-05
Report concrete versions for floating Oracle JDK downloads

### LINKED ISSUE #1212: Oracle and GraalVM floating downloads do not use the actual JDK version
**Description:**

`distribution: oracle` and Oracle `distribution: graalvm` resolve major-only requests through mutable `/latest/` download URLs, but both report the requested major as the resolved release version. For example, a request for Java `21` produces `JavaDownloadRelease.version === "21"` even though the downloaded archive contains a concrete patch/build such as `21.0.x+build`.

Oracle JDK constructs the floating URL and returns `version: range` here:

https://github.com/actions/setup-java/blob/ab597f914a6894678251803485c09f413263b889/src/distributions/oracle/installer.ts#L85-L121

Oracle GraalVM has the same behavior here:

https://github.com/actions/setup-java/blob/ab597f914a6894678251803485c09f413263b889/src/distributions/graalvm/installer.ts#L129-L156

The base installer then compares the existing tool-cache version with that major-only value:

https://github.com/actions/setup-java/blob/ab597f914a6894678251803485c09f413263b889/src/distributions/base-installer.ts#L178-L221

This has two observable consequences:

1. The documented `version` output is not the actual installed Java version; it is only the requested major.
2. `check-latest: true` and the `latest` alias can reuse a stale tool-cache installation after the vendor publishes a newer patch. The remote `/latest/` result is still represented as `21`, which equals the existing tool-cache entry named `21`, so setup-java skips the download even though the floating URL now serves different bytes.

The `floating` flag prevents these releases from entering the resolution cache, but it does not correct the tool-cache comparison or the reported output.

This applies to Oracle JDK and Oracle GraalVM (`distribution: graalvm`). It does not apply to `graalvm-community`, which resolves concrete versions from GitHub release metadata.

**Task version:**

`actions/setup-java@ab597f914a6894678251803485c09f413263b889` (`main`, unreleased v6 development)

**Platform:**
- [x] Ubuntu
- [x] macOS
- [x] Windows

**Runner type:**
- [x] Hosted
- [x] Self-hosted

**Repro steps:**

The output mismatch can be reproduced immediately:

```yaml
jobs:
  actual-version:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        distribution: [oracle, graalvm]
    steps:
      - id: setup
        uses: actions/setup-java@ab597f914a6894678251803485c09f413263b889
        with:
          distribution: ${{ matrix.distribution }}
          java-version: '21'
          check-latest: true
          force-download: true

      - name: Compare reported and installed versions
        shell: bash
        env:
          REPORTED_VERSION: ${{ steps.setup.outputs.version }}
        run: |
          ACTUAL_VERSION=$(sed -n 's/^JAVA_VERSION="\(.*\)"/\1/p' "$JAVA_HOME/release")
          echo "setup-java output: $REPORTED_VERSION"
          echo "JDK release file:  $ACTUAL_VERSION"
          test "$REPORTED_VERSION" = "$ACTUAL_VERSION"
```

The comparison fails because setup-java reports `21`, while `$JAVA_HOME/release` contains the concrete installed version.

The stale `check-latest` path can be covered deterministically in a unit test by resolving two different `/latest/` artifacts for the same major. After the first is stored in the tool cache as `21`, the second resolution is also represented as `21`, and the equality check reuses the first installation.

**Expected behavior:**

- The `version` output identifies the actual installed patch/build.
- Tool-cache entries use an immutable, concrete version.
- `check-latest: true` and `latest` install a newer patch when the vendor's floating artifact changes.

**Actual behavior:**

- The action reports and caches only the major version for these floating downloads.
- A cached installation with that major can be accepted as current without comparing the actual patch/build or artifact identity.

**Suggested fix:**

- Determine the concrete version from authoritative immutable metadata or from the extracted JDK `release` file.
- Return and expose that concrete version after installation.
- Store the tool-cache installation under the concrete normalized version.
- Ensure current-version checks for floating URLs compare the concrete version or immutable artifact identity/checksum rather than only the requested major.
- Add tests that simulate two different `/latest/` artifacts for the same major and verify the second one replaces the first.

jdubois:
Closes #1212

## Summary
- derive the concrete Java patch/build version from the extracted JDK `release` metadata for Oracle JDK and Oracle GraalVM `/latest/` downloads
- force floating requests through remote artifact fingerprinting and never trust source-less tool-cache matches
- bind concrete-version resolution and JDK-cache reuse to the authoritative artifact checksum; when no checksum is available, bypass floating caches and download safely
- preserve exact-version and GraalVM Community behavior while adding regression and scheduled multi-platform E2E output coverage
- regenerate all required `dist/` bundles with the repository CI toolchain

## Validation
- Full Basic validation matrix: Linux, macOS, and Windows (format, lint, build, 1,327 tests, audit): https://github.com/jdubois/setup-java/actions/runs/31010175674
- Final generated `dist/` comparison: https://github.com/jdubois/setup-java/actions/runs/31010359757
- Local npm restore was blocked by the configured package proxy (404 for `brace-expansion@5.0.9`), so validation and bundle generation used the fork workflows as requested
- `bash -n __tests__/verify-java.sh`
- `git diff --check`

### DIFF
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -15,12 +15,6 @@ permissions:
   contents: read
 
 jobs:
-  call-basic-validation:
-    name: Basic validation
-    uses: actions/reusable-workflows/.github/workflows/basic-validation.yml@main
-    with:
-      node-version: '24.x'
-
   call-check-dist:
     name: Check dist/
     uses: actions/reusable-workflows/.github/workflows/check-dist.yml@main
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -15,6 +15,12 @@ permissions:
   contents: read
 
 jobs:
+  call-basic-validation:
+    name: Basic validation
+    uses: actions/reusable-workflows/.github/workflows/basic-validation.yml@main
+    with:
+      node-version: '24.x'
+
   call-check-dist:
     name: Check dist/
     uses: actions/reusable-workflows/.github/workflows/check-dist.yml@main
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -15,37 +15,6 @@ permissions:
   contents: read
 
 jobs:
-  format-artifact:
-    name: Format artifact
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
-      - name: Setup Node.js
-        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
-        with:
-          node-version: '24.x'
-          cache: npm
-      - name: Install dependencies
-        run: npm ci --ignore-scripts
-      - name: Format
-        run: npm run format
-      - name: Upload formatted files
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
-        with:
-          name: formatted-files
-          path: |
-            __tests__/distributors/base-installer.test.ts
-            __tests__/distributors/graalvm-installer.test.ts
-            src/distributions/base-installer.ts
-            src/util.ts
-
-  call-basic-validation:
-    name: Basic validation
-    uses: actions/reusable-workflows/.github/workflows/basic-validation.yml@main
-    with:
-      node-version: '24.x'
-
   call-check-dist:
     name: Check dist/
     uses: actions/reusable-workflows/.github/workflows/check-dist.yml@main
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -15,6 +15,31 @@ permissions:
   contents: read
 
 jobs:
+  format-artifact:
+    name: Format artifact
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - name: Setup Node.js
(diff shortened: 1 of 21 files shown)


brunoborges:
I pushed one commit to this branch (`f733709`) — hope that's alright, happy to drop it if you'd rather handle it separately.

**Context:** I had a parallel implementation of #1212 and compared the two. Yours is the better design — preferring `JAVA_RUNTIME_VERSION` over `JAVA_VERSION` picks up the build number (`21.0.8+12` rather than `21.0.8`), which matches what the other distributions report, and keying the resolution cache on the artifact checksum is much cleaner than what I had.

The one gap I wanted to close is the checksum-less case. Oracle and Oracle GraalVM don't always publish a `.sha256` sibling next to a `/latest/` artifact, and `fetchChecksum` returns `undefined` on a 404. Those releases were excluded from both the resolution cache and the JDK cache, so `cache-jdk` users lose caching entirely for them.

The naive fix — falling back to the URL identity — doesn't work: a `/latest/` URL is a constant string, so the key would never invalidate and a stale JDK would be restored forever, which is exactly the bug this PR fixes. So the commit derives a validator from the HEAD response both installers already make:

- `getArtifactFingerprint(headers)` in `util.ts` — `ETag` when present, otherwise `Last-Modified` + `Content-Length` together (neither alone is enough: one-second granularity, and a same-size rebuild). Returns `undefined` when the response carries no validator.
- `JavaDownloadRelease.fingerprint`, set by the Oracle and GraalVM resolvers for floating releases only. No extra requests.
- `getJdkReleaseIdentity` falls back to the fingerprint before the URL, and a new `hasStableReleaseIdentity()` replaces the four `javaRelease.checksum` gates. A floating release with neither checksum nor fingerprint is still left uncached, as before.

Republishing changes the validator, which changes the key, so a new build is downloaded rather than masked.

41 suites / 1340 tests pass, along with format-check, lint, and a `dist/` rebuild. I checked the guard is load-bearing: reverting `hasStableReleaseIdentity` to checksum-only fails exactly the two new caching tests.

PR #1269, opened 2026-09-12
Support Temurin on Linux RISC-V

yawkat:
**Description:**
Add `riscv64` as a canonical architecture and allow Temurin distributions to use it on Linux.

`setup-java` v5 forwarded the runner architecture to Adoptium and successfully installed Temurin on RISC-V. The centralized platform validation added in v6 currently rejects the same configuration before reaching the Temurin catalog, even though Adoptium publishes Linux `riscv64` JDK and JRE artifacts.

This change:
- adds `riscv64` to the canonical architecture contract;
- enables it only for Temurin on Linux;
- verifies case normalization, platform validation, and Adoptium query pass-through;
- updates the action metadata and compatibility documentation; and
- regenerates `dist/setup/index.js`.

**Related issue:**
https://github.com/yawkat/lz4-java/pull/87

**Check list:**
- [x] Documentation changes are included.
- [x] Tests were added or updated to cover the changes.


yawkat:
@luhenry PTAL as well, since you worked on lz4-java RISC-V support

src/distributions/platform-types.ts
```diff
@@ -35,7 +42,11 @@ export const JAVA_PLATFORM_CAPABILITIES: Record<
 > = {
   [JavaDistribution.Temurin]: {
     platforms: {
-      linux: [...STANDARD_LINUX, {architecture: 'armv7', versionRange: '<18'}],
+      linux: [
+        ...STANDARD_LINUX,
+        'riscv64',
```
luhenry: Shouldn't `riscv64` be added to `STANDARD_LINUX` instead? Its only use is [here](https://github.com/actions/setup-java/pull/1269/changes#diff-8da2075fd8d496f87593d53f449265fea77bc09b55f3e258599d5db1e6c86440R46) AFAICT

__tests__/java-platform-contract.test.ts
```diff
@@ -25,6 +25,7 @@ describe('Java platform capabilities', () => {
     ['aarch64', 'aarch64'],
     ['arm64', 'aarch64'],
     ['ppc64le', 'ppc64le'],
+    ['RISCV64', 'riscv64'],
```
luhenry:
```suggestion
    ['riscv64', 'riscv64'],
```

luhenry (commented):
cc @gdams  @brunoborges @johnoliver

__tests__/java-platform-contract.test.ts
yawkat: Applied in 0c01163—the canonical identity case now uses `['riscv64', 'riscv64']`.

src/distributions/platform-types.ts
yawkat: Agreed. I moved `riscv64` into `STANDARD_LINUX` and regenerated `dist/setup/index.js` in 0c01163.

__tests__/java-platform-contract.test.ts
```diff
@@ -25,6 +25,7 @@ describe('Java platform capabilities', () => {
     ['aarch64', 'aarch64'],
     ['arm64', 'aarch64'],
     ['ppc64le', 'ppc64le'],
+    ['RiScV64', 'riscv64']
```
yawkat: Missing trailing comma here. Without it, the next row is parsed as an index into this one (`['RiScV64', 'riscv64']['s390x', 's390x']`) instead of as a separate `it.each` row. TypeScript rejects the comma-operator index (TS2695), so ts-jest fails to compile this whole test file. Without type-checking it would evaluate to `undefined`, and both the riscv64 and s390x cases would be lost.

```suggestion
    ['RiScV64', 'riscv64'],
```

brunoborges (approved):

